PR #14, opened 2020-02-11
Added cli flag to change overlay_cursor option, defaulting to false

danshick:
Adds the -p/--pointer flag, which allows the user to set the overlay_cursor property in the backend.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -575,6 +575,7 @@ int wayvnc_usage(FILE* stream, int rc)
 "    -o,--output=<name>                        Select output to capture.\n"
 "    -k,--keyboard=<layout>                    Select keyboard layout.\n"
 "    -s,--seat=<name>                          Select seat by name.\n"
+"    -r,--render-cursor                        Enable overlay cursor rendering.\n"
 "    -h,--help                                 Get help (this text).\n"
 "\n";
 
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -575,6 +575,7 @@ int wayvnc_usage(FILE* stream, int rc)
 "    -o,--output=<name>                        Select output to capture.\n"
 "    -k,--keyboard=<layout>                    Select keyboard layout.\n"
 "    -s,--seat=<name>                          Select seat by name.\n"
+"    -p,--pointer=true|false                   Select whether or not to overlay cursor (default to false).\n"
```
any1: Maybe call this render-cursor instead? It shouldn't take an argument.

src/main.c
```diff
@@ -810,6 +817,10 @@ int main(int argc, char* argv[])
 		break;
 	}
 
+  self.capture_backend->overlay_cursor = false;
+  if(overlay_cursor && strncmp(overlay_cursor, "true", 4))
```
any1: I think you mean to say `strcmp(overlay_cursor, "true") == 0`, but that's not needed either.

src/main.c
```diff
@@ -810,6 +817,10 @@ int main(int argc, char* argv[])
 		break;
 	}
 
+  self.capture_backend->overlay_cursor = false;
```
any1: The self structure is initialised to 0 at the top of this function, so this shouldn't be needed. Something strange is going on here.

src/main.c
danshick: Yup, that's obsolete now, but yes I did.

danshick:
How about now?

src/main.c
danshick: I've been conditioned to use strncmp for improperly terminated inputs, but I meant to ! the second term.

src/main.c
danshick: Yeah, it was initialized to false, my poorly written conditional above was faking me out.

any1:
Much better. Can you squash the changes into one commit and force push? Please use imperative wording in the commit message. Something like "Add option to enable cursor overlay rendering".

danshick:
Done, thanks

any1:
Thanks!

PR #17, opened 2020-02-18
Create FAQ.md

danshick:
What do you think about an FAQ page? This might be a good place to keep track of things that aren't common enough to go into the README, but that come up often.


any1:
Thanks!

PR #31, opened 2020-04-05
Fix keyboard shared memory, use memfd/SHM_ANON

myfreeweb:
This fixes the wlroots-on-FreeBSD crash described in https://github.com/swaywm/wlroots/issues/2014 (`write` was returning 0 because the shm fd wasn't `ftruncate`d to size, your TODO check would've helped lol)

and while here: proper shared-memory-fd support with memfd :)

### DIFF
--- a/src/shm.c
+++ b/src/shm.c
@@ -8,6 +8,18 @@
 #include <time.h>
 #include <unistd.h>
 
+#include "config.h"
+
+// Linux with glibc < 2.27 has no wrapper
+#if defined(HAVE_MEMFD) && !defined(HAVE_MEMFD_CREATE)
+#include <sys/syscall.h>
+
+static inline int memfd_create(const char *name, unsigned int flags) {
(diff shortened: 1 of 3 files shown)


src/shm.c
```diff
@@ -8,6 +8,18 @@
...
 #include <unistd.h>
 
+#include "config.h"
+
+// Linux with glibc < 2.27 has no wrapper
+#if defined(HAVE_MEMFD) && !defined(HAVE_MEMFD_CREATE)
+#include <sys/syscall.h>
+
+static inline int memfd_create(const char *name, unsigned int flags) {
+	return syscall(SYS_memfd_create, name, flags);¬
```
jbeich, line 18: Maybe drop `¬` at the end.

any1: Yeah, I was scratching my head about this logic symbol. :)

myfreeweb: well, newline symbol :)

any1:
Thanks!

PR #55, opened 2020-07-16
Unbreak build on BSDs

jbeich:
<endian.h> doesn't exists on DragonFly and FreeBSD. Also, big-endian case has a typo.


any1:
Thanks!

PR #58, opened 2020-07-27
Unbreak build without libgbm

jbeich:
Regressed by 3742dc7144a8 (made `gbm` in meson.build optional). From [error log](https://github.com/any1/wayvnc/files/4984087/wayvnc-0.2.0.log):
```c
../src/buffer.c:25:10: fatal error: 'gbm.h' file not found
#include <gbm.h>
         ^~~~~~~
```
https://github.com/any1/wayvnc/blob/800b0d6cb76ac647615d409e6214861333fd122b/src/main.c#L56-L57

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -22,7 +22,6 @@
 #include <sys/mman.h>
 #include <libdrm/drm_fourcc.h>
 #include <wayland-client.h>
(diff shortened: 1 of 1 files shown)


any1:
Thanks!

PR #59, opened 2020-07-27
Silence -Wunused-function on FreeBSD < 13

jbeich:
Regressed by #31. From [build log](https://github.com/any1/wayvnc/files/4984140/wayvnc-0.2.0.log):
```c
../src/shm.c:35:13: warning: unused function 'randname' [-Wunused-function]
static void randname(char *buf)
            ^
```


any1:
Thanks!

PR #62, opened 2020-08-18
Add openSUSE Tumbleweed installation instruction

agraul:
`wayvnc` is in the official openSUSE Tumbleweed repos and can be installed with a single command.


any1:
Thanks!

PR #66, opened 2020-09-16
Add basic clipboard support

soreau:
Uses wlr-data-control-unstable-v1 protocol to interface with the clipboard
making copy/paste of text possible.

### DIFF
--- a/src/data-control.c
+++ b/src/data-control.c
@@ -20,50 +20,87 @@
 #include <stdbool.h>
 #include <unistd.h>
 #include <assert.h>
+#include <aml.h>
 
 #include "logging.h"
 #include "data-control.h"
 
+struct receive_context {
+	struct data_control* data_control;
+	struct zwlr_data_control_offer_v1* offer;
+	FILE* mem_fp;
+	size_t mem_size;
+	char* mem_data;
+};
+
+static void on_receive(void* handler)
+{
+	struct receive_context* ctx = aml_get_userdata(handler);
+	int fd = aml_get_fd(handler);
+
+	char buf[4096];
+
+	ssize_t ret = read(fd, &buf, sizeof(buf));
+	if (ret > 0) {
+		fwrite(&buf, 1, ret, ctx->mem_fp);
+		return;
+	}
+
+	fclose(ctx->mem_fp);
+
+	if (ctx->mem_size)
+		nvnc_send_cut_text(ctx->data_control->server, ctx->mem_data,
+				ctx->mem_size);
+
+	free(ctx->mem_data);
+	zwlr_data_control_offer_v1_destroy(ctx->offer);
+	aml_stop(aml_get_default(), handler);
+	close(fd);
+}
+
 static void receive_data(void* data,
-	struct zwlr_data_control_offer_v1* zwlr_data_control_offer_v1)
+	struct zwlr_data_control_offer_v1* offer)
 {
 	struct data_control* self = data;
-	char buf[4096];
-	FILE* mem_fp;
-	char* mem_data;
-	size_t mem_size = 0;
 	int pipe_fd[2];
-	int ret;
 
 	if (pipe(pipe_fd) == -1) {
 		log_error("pipe() failed: %m\n");
 		return;
 	}
 
-	zwlr_data_control_offer_v1_receive(zwlr_data_control_offer_v1, self->mime_type, pipe_fd[1]);
+	struct receive_context* ctx = calloc(1, sizeof(*ctx));
+	if (!ctx) {
+		log_error("OOM");
+		close(pipe_fd[0]);
+		close(pipe_fd[1]);
+		return;
+	}
+
+	zwlr_data_control_offer_v1_receive(offer, self->mime_type, pipe_fd[1]);
 	wl_display_flush(self->wl_display);
 	close(pipe_fd[1]);
 
-	mem_fp = open_memstream(&mem_data, &mem_size);
-	if (!mem_fp) {
+	ctx->data_control = self;
+	ctx->offer = offer;
+	ctx->mem_fp = open_memstream(&ctx->mem_data, &ctx->mem_size);
+	if (!ctx->mem_fp) {
+		free(ctx);
+		close(pipe_fd[0]);
 		log_error("open_memstream() failed: %m\n");
+		return;
+	}
+
+	struct aml_handler* handler = aml_handler_new(pipe_fd[0], on_receive,
+			ctx, free);
+	if (!handler) {
+		free(ctx);
 		close(pipe_fd[0]);
 		return;
 	}
 
-	// TODO: Make this asynchronous
-	while ((ret = read(pipe_fd[0], &buf, sizeof(buf))) > 0)
-		if (fwrite(&buf, 1, ret, mem_fp) != ret)
-			break;
-
-	fclose(mem_fp);
-
-	if (mem_size)
-		nvnc_send_cut_text(self->server, mem_data, mem_size);
-
-	free(mem_data);
-	close(pipe_fd[0]);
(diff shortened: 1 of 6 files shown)


src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+#include <unistd.h>
+
+#include "logging.h"
+#include "data-control.h"
+
+static void receive_data(void* data,
+	struct zwlr_data_control_offer_v1 *zwlr_data_control_offer_v1,
+	const char* mime_type)
+{
+	struct data_control* self = (struct data_control*) data;
```
any1, line 30: typecasting is redundant here. This occurs also in other places in this PR.

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+	char c;
+	FILE *mem_fp;
```
any1, line 32: Star is in the wrong side.

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+	close(pipe_fd[1]);
+
+	mem_fp = open_memstream(&mem_data, &mem_size);
+	if (!mem_fp) {
+		log_error("open_memstream() failed: %m\n");
+		close(pipe_fd[0]);
+		return;
+	}
+
+	while (read(pipe_fd[0], &c, 1) > 0)
```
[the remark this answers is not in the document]
any1: Also, what happens if `errno == EINTR`?

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+		self->primary_selection = NULL;
+	}
+	if (!id)
+		return;
+
+	self->primary_selection = id;
+}
+
+static struct zwlr_data_control_device_v1_listener data_control_device_listener = {
+	data_control_device_offer,
```
any1, line 136: prefer named field initialisers.

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+	const char* mime_type)
+{
+	struct data_control* self = (struct data_control*) data;
+
+	if (!strcmp(mime_type, "text/plain"))
+		self->offer = zwlr_data_control_offer_v1;
+}
+
+struct zwlr_data_control_offer_v1_listener data_control_offer_listener = {
+	data_control_offer
```
any1, line 76: also missing name of field

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+	struct zwlr_data_control_source_v1* zwlr_data_control_source_v1,
+	const char* mime_type,
+	int32_t fd)
+{
+	struct data_control* self = (struct data_control*) data;
+	char* d = self->cb_data;
+	size_t len = self->cb_len;
+
+	while (len--)
+		write(fd, d++, 1);
```
any1, line 153: Why not just `write(fd, d, len)`?

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+};
+
+static void
+data_control_source_send(void* data,
+	struct zwlr_data_control_source_v1* zwlr_data_control_source_v1,
+	const char* mime_type,
+	int32_t fd)
+{
+	struct data_control* self = (struct data_control*) data;
+	char* d = self->cb_data;
```
any1, line 149: maybe `assert(d)`?

src/data-control.c
```diff
@@ -0,0 +1,214 @@
...
+	if (self->data_control_source)
+		zwlr_data_control_source_v1_destroy(self->data_control_source);
+
+	self->cb_data = strdup(text);
+	if (!self->cb_data) {
+		log_error("OOM: %m\n");
+		return;
+	}
+	self->cb_len = len;
+	self->data_control_source = zwlr_data_control_manager_v1_create_data_source(self->manager);
```
any1, line 210: There's an error check missing here, I think.

any1 (changes requested):

src/data-control.c
soreau: Check for NULL? Not sure how else to check for errors here.

src/data-control.c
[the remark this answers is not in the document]
soreau: On EINTR, the return value of read() should be <= 0 which means we would have stopped the read loop in this case, right?

src/data-control.c
```diff
@@ -0,0 +1,238 @@
...
+
+	mem_fp = open_memstream(&mem_data, &mem_size);
+	if (!mem_fp) {
+		log_error("open_memstream() failed: %m\n");
+		close(pipe_fd[0]);
+		return;
+	}
+
+	// TODO: Make this asynchronous
+	while ((ret = read(pipe_fd[0], &buf, CHUNK_SIZE)) > 0)
```
any1, line 58: Maybe the check should be `== 0`? I think that's what "EOF" means, which is what the spec says should be used.

There is an example in weston for this here: https://gitlab.freedesktop.org/wayland/weston/-/blob/master/clients/window.c#L3998-4041

Based on this example, it seems that it shouldn't be very hard to make this async.

src/data-control.c
```diff
@@ -0,0 +1,238 @@
...
+		self->cb_data = NULL;
+		return;
+	}
+
+	memcpy(self->cb_data, text, len);
+	self->cb_len = len;
+	zwlr_data_control_source_v1_add_listener(self->data_control_source, &data_control_source_listener, self);
+	zwlr_data_control_source_v1_offer(self->data_control_source, "text/plain;charset=utf-8");
+	zwlr_data_control_device_v1_set_selection(self->data_control_device, self->data_control_source);
+	wl_display_flush(self->wl_display);
```
any1, line 237: Is a flush really needed here?

any1 (approved):

src/data-control.c
soreau: `while ((ret = read(pipe_fd[0], &buf, CHUNK_SIZE)) > 0)` reads until `<= 0` is returned. The code you posted doesn't show anything different than what this PR does than destroy the offer on `read == 0`. In this PR, we simply destroy the offer as the spec indicates [here](https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-data-control-unstable-v1.xml#L152).

soreau: I see that you want to poll the fd and read it when it becomes readable. Doing this requires a lot more work and code, which I am not willing to do.

any1: Ok, let's just leave it as is for now, and I'll make it async before next release. Thanks for your work!

src/data-control.c
```diff
@@ -0,0 +1,235 @@
...
+	self->cb_len = 0;
+}
+
+void data_control_destroy(struct data_control* self)
+{
+	if (self->data_control_source)
+		zwlr_data_control_source_v1_destroy(self->data_control_source);
+	free(self->cb_data);
+
+	zwlr_data_control_manager_v1_destroy(self->manager);
```
any1, line 203: Whoops, there's a double free of self->manager. First here...

src/main.c
```diff
@@ -292,6 +300,8 @@ void wayvnc_destroy(struct wayvnc* self)
 
 	if (self->screencopy.manager)
 		zwlr_screencopy_manager_v1_destroy(self->screencopy.manager);
+	if (self->data_control.manager)
+		zwlr_data_control_manager_v1_destroy(self->data_control.manager);
```
any1, line 304: ... and then here

soreau: Thanks for finding this, should be fixed now. It also should work on sway too now, because I added a blatant hack to workaround the protocol bug.

any1:
I did manage to copy some text from my vnc client after the latest changes, but when I tried to copy text into the vnc client, wayvnc stalled completely. The workaround for the wlroots issue seems racy, so that might be what went wrong for me. Maybe we should focus on solving https://github.com/swaywm/wlroots/issues/2406 first.

any1:
Thanks!

PR #69, opened 2020-09-30
Fixes

MisterDA:
Hi!
Some very small fixes found with PVS-Studio. The analyzer suggest to wrap parameters of the `ALIGN_UP` macro in parentheses, but seeing that it is unused, I removed it.


any1:
It's good you caught that use-after-free. Thanks!

PR #72, opened 2020-10-16
PAM Authentication

NickSica:
Adds barebones PAM authentication.

### DIFF
--- a/include/pam_auth.h
+++ b/include/pam_auth.h
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2019 - 2020 Nicholas Sica
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+ * REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+ * INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+ * LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE
+ * OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
 #pragma once
 
 #include <stdbool.h>
--- a/include/pam_auth.h
+++ b/include/pam_auth.h
@@ -1,15 +1,5 @@
 #pragma once
 
 #include <stdbool.h>
-#include <stdlib.h>
-#include <string.h>
-#include <security/pam_appl.h>
-
-#include "logging.h"
-
-struct credentials {
-	const char* user;
-	const char* password;
-};
 
 bool pam_auth(const char* username, const char* password);
--- a/include/pam_auth.h
+++ b/include/pam_auth.h
@@ -1,0 +1,15 @@
+#pragma once
+
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <security/pam_appl.h>
+
+#include "logging.h"
+
+struct credentials {
+	const char* user;
+	const char* password;
+};
+
+bool pam_auth(const char* username, const char* password);
--- a/meson.build
+++ b/meson.build
@@ -39,6 +39,8 @@ cc = meson.get_compiler('c')
 
 libm = cc.find_library('m', required: false)
 librt = cc.find_library('rt', required: false)
+libpam = cc.find_library('pam', required: get_option('pam'))
+libpam_misc = cc.find_library('pam_misc', required: get_option('pam'))
 
 pixman = dependency('pixman-1')
 gbm = dependency('gbm', required: get_option('screencopy-dmabuf'))
@@ -122,6 +124,12 @@ endif
 
 if gbm.found() and not get_option('screencopy-dmabuf').disabled()
 	config.set('ENABLE_SCREENCOPY_DMABUF', true)
+endif
+
+if libpam.found()
+	dependencies += libpam
+	dependencies += libpam_misc
+	config.set('ENABLE_PAM', true)
 endif
 
 configure_file(
--- a/meson.build
+++ b/meson.build
@@ -40,7 +40,6 @@ cc = meson.get_compiler('c')
 libm = cc.find_library('m', required: false)
 librt = cc.find_library('rt', required: false)
 libpam = cc.find_library('pam', required: get_option('pam'))
-libpam_misc = cc.find_library('pam_misc', required: get_option('pam'))
 
 pixman = dependency('pixman-1')
 gbm = dependency('gbm', required: get_option('screencopy-dmabuf'))
@@ -126,9 +125,9 @@ if gbm.found() and not get_option('screencopy-dmabuf').disabled()
 	config.set('ENABLE_SCREENCOPY_DMABUF', true)
 endif
 
-if libpam.found() and libpam_misc.found()
+if libpam.found()
 	dependencies += libpam
-	dependencies += libpam_misc
+	sources += 'src/pam_auth.c'
 	config.set('ENABLE_PAM', true)
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -126,7 +126,7 @@ if gbm.found() and not get_option('screencopy-dmabuf').disabled()
 	config.set('ENABLE_SCREENCOPY_DMABUF', true)
 endif
 
-if libpam.found()
+if libpam.found() and libpam_misc.found()
 	dependencies += libpam
 	dependencies += libpam_misc
 	config.set('ENABLE_PAM', true)
--- a/src/main.c
+++ b/src/main.c
@@ -34,6 +34,8 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <security/pam_appl.h>
+#include <security/pam_misc.h>
 
 #include "wlr-screencopy-unstable-v1.h"
 #include "wlr-virtual-pointer-unstable-v1.h"
@@ -102,6 +104,11 @@ struct wayvnc {
 	uint32_t damage_area_sum;
 	uint32_t n_frames_captured;
 	uint32_t n_frames_rendered;
+};
+
+struct credentials {
+        const char *user;
+        const char *password;
 };
 
 void wayvnc_exit(struct wayvnc* self);
@@ -461,16 +468,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
 	data_control_to_clipboard(&wayvnc->data_control, text, len);
 }
 
+#ifdef ENABLE_PAM
+static int pam_return_pwd(int num_msg, const struct pam_message **msgm, struct pam_response **response, void *appdata_ptr)
+{
+    struct credentials *cred = (struct credentials *) appdata_ptr;
+    struct pam_response *resp = (struct pam_response *)malloc(sizeof(struct pam_response) * num_msg);
+    for(int i = 0; i < num_msg; i++)
+    {
+        resp[i].resp_retcode = PAM_SUCCESS;
+        switch(msgm[i]->msg_style)
+        {
+        case PAM_TEXT_INFO:
+        case PAM_ERROR_MSG:
+            resp[i].resp = 0;
+            break;
+        case PAM_PROMPT_ECHO_ON:
+            resp[i].resp = cred->user;
+            break;
+        case PAM_PROMPT_ECHO_OFF:
+            resp[i].resp = cred->password;
+            break;
+        }
+    }
+
+    *response = resp;
+    return PAM_SUCCESS;
+}
+#endif
+
 bool on_auth(const char* username, const char* password, void* ud)
 {
+#ifdef ENABLE_PAM
+        struct credentials cred = { strdup(username), strdup(password) };
+        struct pam_conv conv = { &pam_return_pwd, &cred };
+        int result;
+        const char *service = "wayvnc";
+        pam_handle_t *pamh;
+        if((result = pam_start(service, username, &conv, &pamh)) != PAM_SUCCESS)
+        {
+            fprintf(stderr, "ERROR: PAM start failed: %d\n", result);
+            return false;
+        }
+        else if((result = pam_authenticate(pamh, PAM_SILENT|PAM_DISALLOW_NULL_AUTHTOK)) != PAM_SUCCESS)
+        {
+            fprintf(stderr, "PAM authenticate failed: %d\n", result);
+            return false;
+        }
+        else if((result = pam_acct_mgmt(pamh, 0)) != PAM_SUCCESS)
+        {
+            fprintf(stderr, "PAM account management failed: %d\n", result);
+            return false;
+        }
+#else
 	struct wayvnc* self = ud;
-
-	if (strcmp(username, self->cfg.username) != 0)
+        if (strcmp(username, self->cfg.username) != 0)
 		return false;
 
 	if (strcmp(password, self->cfg.password) != 0)
 		return false;
-
+#endif
 	return true;
 }
 
@@ -717,7 +773,7 @@ int check_cfg_sanity(struct cfg* cfg)
 			log_error("Authentication enabled, but missing private_key_file\n");
 			rc = -1;
 		}
-
+#ifndef ENABLE_PAM
 		if (!cfg->username) {
 			log_error("Authentication enabled, but missing username\n");
 			rc = -1;
@@ -727,7 +783,7 @@ int check_cfg_sanity(struct cfg* cfg)
 			log_error("Authentication enabled, but missing password\n");
 			rc = -1;
 		}
-
+#endif
 		return rc;
 	}
 
--- a/src/main.c
+++ b/src/main.c
@@ -108,8 +108,8 @@ struct wayvnc {
 
 #ifdef ENABLE_PAM
 struct credentials {
-        const char *user;
-        const char *password;
+	const char* user;
+	const char* password;
 };
 #endif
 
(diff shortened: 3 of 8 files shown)


meson.build
```diff
@@ -124,6 +126,12 @@ if gbm.found() and not get_option('screencopy-dmabuf').disabled()
 	config.set('ENABLE_SCREENCOPY_DMABUF', true)
 endif
 
+if libpam.found()
```
any1, line 129: Don't you need to check for `libpam_misc` too?

src/main.c
```diff
@@ -34,6 +34,8 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <security/pam_appl.h>
```
any1, line 37: Missing `#ifdef ENABLE_PAM`

src/main.c
```diff
@@ -104,6 +106,13 @@ struct wayvnc {
 	uint32_t n_frames_rendered;
 };
 
+#ifdef ENABLE_PAM
+struct credentials {
+        const char *user;
```
any1, line 111: Style: This project uses tabs for indentation. The `*` should be on the left side.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
 	data_control_to_clipboard(&wayvnc->data_control, text, len);
 }
 
+#ifdef ENABLE_PAM
+static int pam_return_pwd(int num_msg, const struct pam_message **msgm, struct pam_response **response, void *appdata_ptr)
```
any1, line 474: Style: Stars should be on the left side and the line length should not exceed 80 characters.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+{
+    struct credentials *cred = (struct credentials *) appdata_ptr;
```
any1, line 476: Typecasting is redundant for void pointers.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+    struct pam_response *resp = (struct pam_response *)malloc(sizeof(struct pam_response) * num_msg);
```
any1, line 477: Typecasting is not needed for malloc. Prefer `calloc(sizeof(*resp), num_msg)`.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+    for(int i = 0; i < num_msg; i++)
```
any1, line 478: Style: Use tabs, `{` should be inline except in function declarations. There should be a space between `for` and `(`; i.e. `for (int i = 0; i < num_msg; i++) {`

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+    {
+        resp[i].resp_retcode = PAM_SUCCESS;
+        switch(msgm[i]->msg_style)
```
any1, line 481: Ditto

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+    }
+
+    *response = resp;
+    return PAM_SUCCESS;
+}
+#endif
+
 bool on_auth(const char* username, const char* password, void* ud)
 {
+#ifdef ENABLE_PAM
```
any1, line 503: Enabling PAM should be a config and/or command line option. If a package is built and released with PAM enabled it will break authentication for current users if done like this, so it needs to be an opt-in.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+        struct credentials cred = { strdup(username), strdup(password) };
+        struct pam_conv conv = { &pam_return_pwd, &cred };
+        int result;
+        const char *service = "wayvnc";
+        pam_handle_t *pamh;
+        if((result = pam_start(service, username, &conv, &pamh)) != PAM_SUCCESS)
```
any1, line 509: I think you need to call `pam_end()` somewhere. Otherwise the handle is leaked.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+        {
+            fprintf(stderr, "ERROR: PAM start failed: %d\n", result);
+            return false;
+        }
+        else if((result = pam_authenticate(pamh, PAM_SILENT|PAM_DISALLOW_NULL_AUTHTOK)) != PAM_SUCCESS)
```
any1, line 514: `else` is redundant, due to `return` above.

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+        {
+            fprintf(stderr, "PAM authenticate failed: %d\n", result);
+            return false;
+        }
+        else if((result = pam_acct_mgmt(pamh, 0)) != PAM_SUCCESS)
```
any1, line 519: ditto

src/main.c
```diff
@@ -461,16 +470,65 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+        {
+            fprintf(stderr, "PAM account management failed: %d\n", result);
```
any1, line 521: There's a function in wayvnc called `log_error` which serves this purpose.

any1 (changes requested):
How does this work? Can any user log in with their own credentials, or do they need to be a part of some group or something?

NickSica:
> How does this work? Can any user log in with their own credentials, or do they need to be a part of some group or something?
The wayvnc.pam file needs to be placed in /etc/pam.d/ and defines how authentication for the service is done. This passes the authentication request onto PAM. In theory you could restrict it to groups or pass it to something like LDAP and only need to change the config to do so but my barebones version just authenticates the user.

I'll fix the style changes, I already fixed most of the other errors but forgot to push it to the pull request last night.

meson.build
NickSica: I'm not exactly sure if one can ever come without the other, but I'll check for both.

src/main.c
NickSica: Trying to wrap it with the ifdef seemed to break other stuff even though the other stuff was wrapped as well.

NickSica:
Since it is in the config file now, I didn't know if you wanted to just get rid of the ENABLE_PAM in the build config or not. People would have to have pam on their system to build it, but I think most do already.

src/main.c
any1: You probably need to include at a later point where "config.h" has been included. It should be included here explicitly, but I guess it's included by some other header.

src/main.c
```diff
@@ -461,16 +470,74 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
 	data_control_to_clipboard(&wayvnc->data_control, text, len);
 }
 
+#ifdef ENABLE_PAM
+static int pam_return_pwd(int num_msg, const struct pam_message** msgm,
+			  struct pam_response** response, void* appdata_ptr)
+{
+    struct credentials* cred = appdata_ptr;
```
any1, line 477: There are still some spaces in this PR which should be tabs.

src/main.c
```diff
@@ -461,16 +470,74 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+    struct pam_response* resp = calloc(sizeof(*response), num_msg);
+    for (int i = 0; i < num_msg; i++) {
+	resp[i].resp_retcode = PAM_SUCCESS;
+	switch(msgm[i]->msg_style)
```
any1, line 481: Style: `{` should be inline.

src/main.c
```diff
@@ -461,16 +470,74 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
 	data_control_to_clipboard(&wayvnc->data_control, text, len);
 }
 
+#ifdef ENABLE_PAM
+static int pam_return_pwd(int num_msg, const struct pam_message** msgm,
+			  struct pam_response** response, void* appdata_ptr)
+{
+    struct credentials* cred = appdata_ptr;
+    struct pam_response* resp = calloc(sizeof(*response), num_msg);
```
any1, line 478: Where is this freed?

src/main.c
```diff
@@ -461,16 +470,74 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+    return PAM_SUCCESS;
+}
+#endif
+
 bool on_auth(const char* username, const char* password, void* ud)
 {
 	struct wayvnc* self = ud;
+#ifdef ENABLE_PAM
+	if (self->cfg.enable_pam) {
+	    struct credentials cred = { strdup(username), strdup(password) };
```
any1, line 506: Where are these freed?

It would be nicer to have this code block within this `if` extracted to a function.

src/main.c
```diff
@@ -461,16 +470,74 @@ static void on_client_cut_text(struct nvnc* server, const char* text, uint32_t l
...
+	    }
+	}
+	else {
+#endif
+	    if (strcmp(username, self->cfg.username) != 0)
 		return false;
 
+	    if (strcmp(password, self->cfg.password) != 0)
+		return false;
+#ifdef ENABLE_PAM
```
any1, line 538: You can avoid this `#ifdef` by doing an early return in the pam case. I strongly prefer early returns in general.

any1 (changes requested):

src/main.c
NickSica: this gets assigned to pam_response which gets handled further by pam I think. I'm guessing a pam internal function frees it but I'm not 100% sure. I can't free it there because it is used by pam internally.

NickSica:
I don't know if you get notified of new commits, but I pushed the changes you requested.

any1:
> I don't know if you get notified of new commits, but I pushed the changes you requested.

Yeah, sorry, I spent half a day last weekend trying to figure out if there was some alternate route to using the conversation function like this because it feels like a dirty hack and not at all how it was intended to be used.

There's also an issue with this that a failed login will cause a delay in the main loop and this halts all events for the duration of the delay, which can be a few seconds. In the best case, this causes frustration and in the worst case, it makes a DoS attack very easy. I'm afraid that if we want to go this hacky route, then we need to run authentication in a worker thread.

NickSica:
> > I don't know if you get notified of new commits, but I pushed the changes you requested.
> 
> Yeah, sorry, I spent half a day last weekend trying to figure out if there was some alternate route to using the conversation function like this because it feels like a dirty hack and not at all how it was intended to be used.
> 
> There's also an issue with this that a failed login will cause a delay in the main loop and this halts all events for the duration of the delay, which can be a few seconds. In the best case, this causes frustration and in the worst case, it makes a DoS attack very easy. I'm afraid that if we want to go this hacky route, then we need to run authentication in a worker thread.

Hey no problem! Really? I kind of used tigervnc code as a reference for this implementation so it's interesting it's hacky.

any1:
I guess this is as good as it gets with the PAM API being the way it is. Can you move this into its own source file? At least that way, we can keep things a little bit neater.

The delay issue that I mentioned can be worked around by using the `nodelay` option in the PAM config file, but in that case, it might be a good idea to add a timer that inhibits login attempts to make brute force attacks harder.

NickSica:
> I guess this is as good as it gets with the PAM API being the way it is. Can you move this into its own source file? At least that way, we can keep things a little bit neater.
> 
> The delay issue that I mentioned can be worked around by using the `nodelay` option in the PAM config file, but in that case, it might be a good idea to add a timer that inhibits login attempts to make brute force attacks harder.

Ok I made the changes requested, let me know if there are any issues.

src/pam_auth.c
```diff
@@ -0,0 +1,58 @@
...
+		log_error("PAM authenticate failed: %d\n", result);
+		return false;
+	}
+
+	if ((result = pam_acct_mgmt(pamh, 0)) != PAM_SUCCESS) {
+		log_error("PAM account management failed: %d\n", result);
+		return false;
+	}
+
+	if ((result = pam_end(pamh, result)) != PAM_SUCCESS) {
```
any1, line 53: You don't need to check the result of this. `pam_end` only returns error if `pamh == NULL`, which should not be the case.

src/pam_auth.c
```diff
@@ -0,0 +1,58 @@
...
+}
+
+bool pam_auth(const char* username, const char* password)
+{
+	struct credentials cred = { username, password };
+	struct pam_conv conv = { &pam_return_pwd, &cred };
+	int result;
+	const char* service = "wayvnc";
+	pam_handle_t* pamh;
+	if ((result = pam_start(service, username, &conv, &pamh)) != PAM_SUCCESS) {
```
any1, line 38: It's nicer to not mix the assignment and the error check. I'd rather like to see:
```
result = pam_start(...)
if (result != PAM_SUCCESS) {...
```
this comment also applies for the error checks below.

any1 (changes requested):
Modifying .gitignore should be a separate commit. See https://chris.beams.io/posts/git-commit/

src/pam_auth.c
```diff
@@ -0,0 +1,58 @@
...
+	struct pam_response* resp = calloc(sizeof(*response), num_msg);
+	for (int i = 0; i < num_msg; i++) {
+		resp[i].resp_retcode = PAM_SUCCESS;
+		switch(msgm[i]->msg_style) {
+		case PAM_TEXT_INFO:
+		case PAM_ERROR_MSG:
+			resp[i].resp = 0;
+			break;
+		case PAM_PROMPT_ECHO_ON:
+			resp[i].resp = strdup(cred->user);
```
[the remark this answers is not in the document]
NickSica: I figured, but I wanted to be thorough. I'll take out the extra stuff though.

src/pam_auth.c
```diff
@@ -0,0 +1,58 @@
...
+			break;
+		case PAM_PROMPT_ECHO_OFF:
+			resp[i].resp = strdup(cred->password);
+			break;
+		default:
+			free(resp);
```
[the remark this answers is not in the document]
NickSica: I read somewhere the response structure is freed by PAM in a different function if authentication is successful. If it isn't we have to free it which is what is happening here. That might be a different version of PAM though and not the linux version. Let me play around with this to see.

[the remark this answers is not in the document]
any1: Yes, that's what I read too. What I'm commenting on is that if there is an error that occurs, it's not enough to do `free(resp)`, you also need to free `resp[i].resp` for each `i`.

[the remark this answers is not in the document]
NickSica: Oh gotcha, I'll add that in.

NickSica:
> Modifying .gitignore should be a separate commit. See https://chris.beams.io/posts/git-commit

> Modifying .gitignore should be a separate commit. See https://chris.beams.io/posts/git-commit/

What would be the best way to split these using git? My git knowledge is very basic

any1:
> > Modifying .gitignore should be a separate commit. See [chris.beams.io/posts/git-commit](https://chris.beams.io/posts/git-commit)
> 
> What would be the best way to split these using git? My git knowledge is very basic

We can do that at the end with `git rebase --interactive`. Then we'll squash the commits that add pam into one and split out the .gitignore one.

NickSica:
> > > Modifying .gitignore should be a separate commit. See [chris.beams.io/posts/git-commit](https://chris.beams.io/posts/git-commit)
> > 
> > 
> > What would be the best way to split these using git? My git knowledge is very basic
> 
> We can do that at the end with `git rebase --interactive`. Then we'll squash the commits that add pam into one and split out the .gitignore one.

Ok, thank you for the patience. This is the first open source project I've contributed to.

include/pam_auth.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <stdbool.h>
+#include <stdlib.h>
```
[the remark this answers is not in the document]
NickSica: I get implicit declaration warnings when I remove the includes that you recommended to remove.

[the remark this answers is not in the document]
NickSica: Nevermind I think I misunderstood the comments you wanted me to move everything except stdbool.h into the .c file. I moved stdbool.h into the .c and removed the others.

NickSica:
Ok that should resolve all your comments

src/pam_auth.c
```diff
@@ -0,0 +1,77 @@
...
+bool pam_auth(const char* username, const char* password)
+{
+	struct credentials cred = { username, password };
+	struct pam_conv conv = { &pam_return_pwd, &cred };
+	const char* service = "wayvnc";
+	pam_handle_t* pamh;
+	int result = pam_start(service, username, &conv, &pamh);
+	if (result != PAM_SUCCESS) {
+		log_error("ERROR: PAM start failed: %d\n", result);
+		goto error;
```
any1, line 50: Here you can just return false;

src/pam_auth.c
```diff
@@ -0,0 +1,77 @@
...
+	if (result != PAM_SUCCESS) {
+		log_error("PAM account management failed: %d\n", result);
+		goto error;
+	}
+
+	pam_end(pamh, result);
+	return true;
+
+error:
+	free(pamh);
```
any1, line 69: `pam_end` is used to free `pamh`.

src/pam_auth.c
```diff
@@ -0,0 +1,77 @@
...
+		log_error("PAM authenticate failed: %d\n", result);
+		goto error;
+	}
+
+	result = pam_acct_mgmt(pamh, 0); 
+	if (result != PAM_SUCCESS) {
+		log_error("PAM account management failed: %d\n", result);
+		goto error;
+	}
+
```
any1, line 64: Put `error:` here and instead of `return true` do `return result == PAM_SUCCESS`?

src/pam_auth.c
```diff
@@ -0,0 +1,77 @@
...
+	pam_end(pamh, result);
+	return true;
+
+error:
+	free(pamh);
+	return false;
+}
+
```
any1, line 72: Hehe, there's a lot of trailing newlines at the end of this file.

wayvnc.pam
```diff
@@ -0,0 +1,2 @@
+auth    required pam_unix.so nodelay deny=3 unlock_time=600
```
any1, line 1: This seems like a good solution.

any1 (commented):
Almost there!

src/pam_auth.c
NickSica: Oh sorry I've been putting off adding something to auto-trim them in my editor

src/pam_auth.c
NickSica: I'm not 100% sure if I understood this comment correctly but check the latest.

any1: I meant to put the `error:` tag above the call to `pam_end`.

NickSica: whoops! I'll fix that haha

src/pam_auth.c
```diff
@@ -0,0 +1,68 @@
...
+bool pam_auth(const char* username, const char* password)
+{
+	struct credentials cred = { username, password };
+	struct pam_conv conv = { &pam_return_pwd, &cred };
+	const char* service = "wayvnc";
+	pam_handle_t* pamh;
+	int result = pam_start(service, username, &conv, &pamh);
+	if (result != PAM_SUCCESS) {
+		log_error("ERROR: PAM start failed: %d\n", result);
```
any1, line 49: Sorry, I keep noticing new things. It's probably nicer to give the user a readable error message. `pam_strerror(result)` should do that.

LGTM otherwise. Do you want to have a go a rebasing?

NickSica: Sure haha how do you wanna do this?

NickSica: If I attempt to pass pam_strerror to wayvnc's log error function, there are some issues. Should I print directly to stderr?

any1: Works fine on my system...
```
[andri@galago wayvnc]$ git diff
diff --git a/src/pam_auth.c b/src/pam_auth.c
index 4ff69de..3d4b595 100644
--- a/src/pam_auth.c
+++ b/src/pam_auth.c
@@ -46,7 +46,7 @@ bool pam_auth(const char* username, const char* password)
        pam_handle_t* pamh;
        int result = pam_start(service, username, &conv, &pamh);
        if (result != PAM_SUCCESS) {
-               log_error("ERROR: PAM start failed: %d\n", result);
+               log_error("ERROR: PAM start failed: %s\n", pam_strerror(pamh, result));
                return false;
        }
```

NickSica: There we go should be good now

NickSica:
This should be good to go correct? Sorry I'm not sure what you get notified about.

any1:
Ahh, there's one more thing. The new files are missing license & copyright.

NickSica:
> Ahh, there's one more thing. The new files are missing license & copyright.

Should I copy the copyright you have in the files and replace your name with mine?

any1:
> > Ahh, there's one more thing. The new files are missing license & copyright.
> 
> Should I copy the copyright you have in the files and replace your name with mine?

Yes, that will do

NickSica:
> > > Ahh, there's one more thing. The new files are missing license & copyright.
> > 
> > 
> > Should I copy the copyright you have in the files and replace your name with mine?
> 
> Yes, that will do

Done

any1 (approved):

NickSica:
Do you have to do all the rebasing or do I do it?

any1:
I did a behind-the-scenes merge. See https://github.com/any1/wayvnc/commit/6a73f293fc60e9297695f196ba709cc3e60e12af

Thanks!

any1:
By the way, in case you're looking for more things to do, it would be nice to have PAM usage documented in the man page.

NickSica:
If I get more free time, I'd be glad to help out. School is keeping me pretty busy as is, though.

Issue #71, opened 2020-10-15
Some modifier keys are not released


Flakebi:
Thanks a lot for this project, it’s so useful!
So, I’m using a weird keyboard layout with multiple layers and all kind of functionality and keys: https://neo-layout.org
It’s the *de* layout in the *neo* variant. I don’t think there is a layout that has these keys, so I need to edit https://github.com/any1/wayvnc/blob/master/src/keyboard.c#L161 and set `.variant = "neo"` to get all keys working (like you described in #61).

When I press the modifier key to get to layer 4 (called Mod 4 on the neo website, it reads “Alt Gr” on my keyboard), I can use the arrow keys as expected. However, after I release the modifier key, the layer is still active and I can’t type characters. Sometimes I can reset this by pressing both Mod 4 keys at the same time twice (which locks and unlocks the layer).

wayvnc log:
```python
# Press Mod 4
DEBUG: symbol=ISO_Level5_Shift level=0 code=LSGT pressed
# Press and release A (which is Down in layer 4; on us layout, this is the D key)
DEBUG: symbol=Down level=0 code=DOWN pressed
DEBUG: symbol=Down level=0 code=DOWN released
# Release Mod 4 (not registered)
# Press A again (D key on us layout), still registered as Down though it *should be registered as A*
DEBUG: Failed to match level on symbol: a
DEBUG: symbol=a level=0 code=AC03 pressed
DEBUG: Failed to match level on symbol: a
DEBUG: symbol=a level=0 code=AC03 released
DEBUG: Failed to match level on symbol: a
DEBUG: symbol=a level=0 code=AC03 pressed
DEBUG: Failed to match level on symbol: a
DEBUG: symbol=a level=0 code=AC03 released
DEBUG: Failed to match level on symbol: a
DEBUG: symbol=a level=0 code=AC03 pressed
DEBUG: Failed to match level on symbol: a
DEBUG: symbol=a level=0 code=AC03 released
# Press and release both Mod 4 keys to lock and unlock layer 4
DEBUG: symbol=ISO_Level5_Shift level=0 code=MDSW pressed
DEBUG: symbol=ISO_Level5_Lock level=4 code=LSGT released
DEBUG: symbol=ISO_Level5_Shift level=0 code=MDSW released
DEBUG: symbol=ISO_Level5_Shift level=0 code=LSGT pressed
DEBUG: symbol=ISO_Level5_Lock level=4 code=LSGT released
# Press and release A again (this time correctly not interpreted as Down)
[...]

PR #73, opened 2020-10-16
Fix de-neo layout

Flakebi:
- Add possibility to specify layout variant with --keyboard-variant
- Add ISO_Level5_Shift and ISO_Level5_Lock to modifier keys

Fixes most of #71, thanks for your guidance!

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -98,6 +98,7 @@ struct wayvnc {
 	struct pixman_region16 current_damage;
 
 	const char* kb_layout;
+	const char* kb_variant;
 
 	uint32_t damage_area_sum;
 	uint32_t n_frames_captured;
@@ -684,7 +685,7 @@ int wayvnc_usage(FILE* stream, int rc)
 "\n"
 "    -C,--config=<path>                        Select a config file.\n"
 "    -o,--output=<name>                        Select output to capture.\n"
-"    -k,--keyboard=<layout>                    Select keyboard layout.\n"
+"    -k,--keyboard=<layout>[-<variant>]        Select keyboard layout with an optional variant.\n"
 "    -s,--seat=<name>                          Select seat by name.\n"
 "    -r,--render-cursor                        Enable overlay cursor rendering.\n"
 "    -f,--max-fps=<fps>                        Set the rate limit (default 30).\n"
@@ -762,6 +763,18 @@ static void start_performance_ticker(struct wayvnc* self)
 	aml_unref(ticker);
 }
 
+void parse_keyboard_option(struct wayvnc* self, char* arg)
+{
+	// Find optional variant, separated by -
+	char* index = strchr(arg, '-');
+	if (index != NULL) {
+		self->kb_variant = index + 1;
+		// layout needs to be 0-terminated, replace the - by 0
+		*index = 0;
+	}
+	self->kb_layout = arg;
+}
+
 int show_version(void)
 {
 	printf("wayvnc: %s\n", wayvnc_version);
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -685,6 +686,7 @@ int wayvnc_usage(FILE* stream, int rc)
 "    -C,--config=<path>                        Select a config file.\n"
 "    -o,--output=<name>                        Select output to capture.\n"
 "    -k,--keyboard=<layout>                    Select keyboard layout.\n"
+"    --keyboard-variant=<variant>              Select a variant of the keyboard layout.\n"
```
any1, line 689: Hmm, maybe have short option too? I don't like 'v' for this because most often it means "verbose" or "version". Any other character that is not taken will do, I suppose.

Another way of doing this would be to include it in the keyboard option so you could say `--keyboard=de` or `--keyboard=de,neo` and the help text would say something like `--keyboard=<layout>[,variant]`. But this also requires a little extra parsing code.

Flakebi: I like the idea of including the variant in the keyboard option. I implemented it is `--keyboard=<layout>[-<variant>]`.

src/main.c
```diff
@@ -814,9 +815,22 @@ int main(int argc, char* argv[])
...
 			output_name = optarg;
 			break;
-		case 'k':
-			self.kb_layout = optarg;
+		case 'k': {
+			// Find optional variant, separated by -
+			char* index = strchr(optarg, '-');
+			if (index != NULL) {
+				self.kb_variant = index + 1;
+				// layout needs to be 0-terminated but optarg may point to a
```
any1, line 823: According to the `man getopt_long`, `optarg` is defined as `extern char* optarg`. I don't mind modifying arguments.

src/main.c
```diff
@@ -814,9 +815,22 @@ int main(int argc, char* argv[])
...
+				// const string, so need to copy
+				char* layout = malloc(index - optarg + 1);
```
any1, line 825: I don't see a `free()` for this anywhere.

src/main.c
Flakebi: If I read the source of getopt right, *optarg` can point to argv, which is a `const char*`, that’s why I copy it.

src/main.c
Flakebi: This would need to be freed at the end of `main`, however, I don’t save if the layout string was alloced or not, so I would have to store a flag for that in the wayvnc struct. Should I do this?

src/main.c
any1: But argv isn't `const char*`. You're allowed to change it. It belongs to the process's memory. :)

Flakebi: That is a legitimate argument and makes it much easier!

src/main.c
any1: If I absolutely needed to `malloc` here, I'd declare the variable at the start of main and assign NULL to it. `free(NULL)` is a no-op. However, a static buffer of a reasonable size would be better. The name of a keyboard layout or variant is unlikely to be longer than 10 characters, so a good length for the buffer would be two times 10, aligned up to nearest multiple of 2, which gives 32.

Benign memory leaks like this one are annoying because they get in the way when trying to find malignant memory leaks with `valgrind`.

src/main.c
any1: Looks good now. Maybe extract the parsing into a function? You could call it something like `parse_keyboard_option`.

any1:
Thanks!

PR #79, opened 2020-12-19
Don't free the front buffer if screencopy fails

### LINKED ISSUE #65: segfault when attached display changes dpms state
I'm encountering a segfault when starting wayvnc from within an ssh session, even with WAYLAND_DISPLAY and SWAYSOCK set. wayvnc is 0.2.0, neatvnc is 0.2.0, and aml is 0.1.0.

I expect this isn't hard to reproduce, but if it is please let me know.

Arnavion:
It is possible for `screencopy_failed` to be called without
the front buffer ever being set, such as when the output is dpms-off.

Fixes #65

### DIFF
--- a/src/screencopy.c
+++ b/src/screencopy.c
@@ -191,7 +191,8 @@ static void screencopy_failed(void* data,
 
 	screencopy__stop(self);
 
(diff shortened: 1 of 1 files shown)


any1:
Thanks!

PR #95, opened 2021-04-04
Support UNIX domain sockets

r-c-f:
Makes use of the functionality added in
https://github.com/any1/neatvnc/pull/49 to support UNIX domain sockets
with a command line flag. This is what I was using to test that pull request, so it makes sense to open this now that it has been accepted. 

### DIFF
--- a/wayvnc.scd
+++ b/wayvnc.scd
@@ -31,6 +31,10 @@ wayvnc - A VNC server for wlroots based Wayland compositors.
 *-p, --show-performance*
 	Show performance counters.
 
+*-u, --unix-socket*
+	Create a UNIX domain socket instead of TCP, treating the address as a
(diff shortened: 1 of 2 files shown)


wayvnc.scd
```diff
@@ -31,6 +31,10 @@ wayvnc - A VNC server for wlroots based Wayland compositors.
 *-p, --show-performance*
 	Show performance counters.
 
+*-u, --unix-socket*
+	Creat a UNIX domain socket instead of TCP, treating the address as a
```
any1, line 35: There's a typo in the first word here. Looks good otherwise.

any1:
LGTM. Can you consolidate into single commit?

any1:
Thanks!

PR #118, opened 2021-12-13
Add usage example to README.md

mb720:
Here's a small usage example that I would loved to see available when I started using wayvnc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,14 @@ user authentication enabled. The best way to protect your VNC connection is to
 use SSH tunneling while listening on localhost, but users can also be
 authenticated when connecting to wayvnc.
 
+### Connect to running Wayland session
+Let's say you want to control the host at `192.168.178.28` which is already up and running Wayland:
+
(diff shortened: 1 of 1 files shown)


lack:
I like the sentiment of this patch, but I think there are 2 things that would make it better:

- Move it into `FAQ.md`
- If you're setting up ssh anyway, it's much safer to tunnel your VNC traffic through SSH using `ssh -L 5900:localhost:5900 $user@$host wayvnc`, and then connect your client to localhost:5900

lack (changes requested):
I like the sentiment of this patch, but I think there are 2 things that would make it better:

- Move it into `FAQ.md`
- If you're setting up ssh anyway, it's much safer to tunnel your VNC traffic through SSH using `ssh -L 5900:localhost:5900 $user@$host wayvnc`, and then connect your client to localhost:5900

bangert:
@any1 i guess this one can be closed?

PR #126, opened 2022-01-26
Function without virtual keyboard, mouse protocol

Consolatis:
Some compositors have a working screencopy protocol implementation but
are missing virtual keyboard and/or virtual mouse protocols. On those
compositors, don't just outright fail, instead work as view-only mode.
This is useful for using a different system (like a tablet) as
additional screen without the system providing any kind of input.

I did not dive deep into wayvnc and its possible this patch is not complete.
It has been tested by using Remmina and some Android client to send input and wayvnc did not crash.
This patch started as a quick hack to get an additional remote screen working with [labwc](https://github.com/labwc/labwc) but I thought it may be relevant to others as well. Feel free to just close this PR if it doesn't match your expectations.

 I have read and understood CONTRIBUTING.md ~~and its associated documents~~.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -283,12 +283,16 @@ void wayvnc_destroy(struct wayvnc* self)
 
 	wl_shm_destroy(wl_shm);
 
-	zwp_virtual_keyboard_v1_destroy(self->keyboard_backend.virtual_keyboard);
-	zwp_virtual_keyboard_manager_v1_destroy(self->keyboard_manager);
-	keyboard_destroy(&self->keyboard_backend);
+	if (self->keyboard_manager) {
+		zwp_virtual_keyboard_v1_destroy(self->keyboard_backend.virtual_keyboard);
+		zwp_virtual_keyboard_manager_v1_destroy(self->keyboard_manager);
+		keyboard_destroy(&self->keyboard_backend);
+	}
 
-	zwlr_virtual_pointer_manager_v1_destroy(self->pointer_manager);
-	pointer_destroy(&self->pointer_backend);
+	if (self->pointer_manager) {
+		zwlr_virtual_pointer_manager_v1_destroy(self->pointer_manager);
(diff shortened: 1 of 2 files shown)


any1:
At a glance, this looks harmless.

I think that the expected behaviour is for wayvnc to fail if there are not inputs for it. I.e. people might be confused if they don't read the output on the command line (many don't). It's probably better to have a flag for disabling inputs. That way, it's more obvious when inputs fail and we get an extra feature that allows people to disable inputs for a "view only mode".

Consolatis:
> It's probably better to have a flag for disabling inputs.

- added a `--disable-input` flag
- if not given wayvnc fails as it did before
- extended the error logs when there is no support for the virtual keyboard or mouse protocol

any1 (changes requested):
The callback for pointer events is set here: https://github.com/any1/wayvnc/pull/126/files#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176L507-L508

If `pointer_manager != NULL` and `disable-input` is enabled, this will crash. The simplest fix is to check for `self->pointer_backend.pointer` rather than `self->pointer_manager`. That's the same as is done below for the keyboard.

Maybe cut-text should also be disabled with inputs for security reasons?

Consolatis:
> The simplest fix is to check for self->pointer_backend.pointer rather than self->pointer_manager

Done. Also added the commit message as the scope of the commit has changed from just ignoring the errors into actually disabling all remote input. I can't really test the feature on a compositor which provides virtual keyboard + mouse protocols right now.

> Maybe cut-text should also be disabled with inputs for security reasons?

Assuming this means VNC clients are able to set the clipboard selection I'd say it makes sense to disable that feature when wayvnc has been started with `--disable-input`.

any1:
Merged. Thanks!

PR #127, opened 2022-01-27
Add meson subproject wraps for aml and neatvnc

Consolatis:
Makes compiling as easy as:
meson build_dir && meson compile -C build_dir

### DIFF
--- a/subprojects/aml.wrap
+++ b/subprojects/aml.wrap
@@ -1,0 +1,3 @@
+[wrap-git]
+url = https://github.com/any1/aml.git
+revision = master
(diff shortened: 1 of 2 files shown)


eli-schwartz:
https://github.com/any1/wayvnc/blob/3b26a43b97722be451ac851597e36ecd11c4ba8a/meson.build#L52-L69

The use of subprojects as preferred with fallback to system dependencies means that adding a wrap file actually breaks the ability to build with system deps. This may be the reason that they aren't already wrap files.

I would advise nuking all this code though. See https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section

`dependency('foo')` should be enough, and simply use a suitable wrap file that specifies a `[provide]` section. Meson will natively do subproject fallback, and it can even be configured to prefer subprojects using `--wrap-mode=forcefallback`.

PR #129, opened 2022-02-10
When remote input is disabled, also disable clipboard management

yobert:
Hello! I think a good feature would be that when remote input is disabled, also disable the clipboard function.

My reasoning is that a common use case for input being disabled is you are showing your screen to a large or untrusted audience, and many password managers use the clipboard as a medium for passwords. Someone might not realize that even though they have remote input disabled, their password may become visible to a guest who pulls up the clipboard tool.

Please forgive my implementation--- I am definitely not a C++ expert. Also thank you for making such awesome software.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -98,6 +98,8 @@ struct wayvnc {
 
 	uint32_t damage_area_sum;
 	uint32_t n_frames_captured;
+
+	bool disable_input;
 };
 
 void wayvnc_exit(struct wayvnc* self);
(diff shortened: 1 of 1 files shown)


src/data-control.c
```diff
@@ -224,6 +224,10 @@ struct zwlr_data_control_source_v1_listener data_control_source_listener = {
 };
 
 static struct zwlr_data_control_source_v1* set_selection(struct data_control* self, bool primary) {
+	if(!self->manager) {
```
any1, line 227: This doesn't do anything if `data_control` isn't initialised.

yobert: Cool! Okay I took out that check and force pushed to update the PR.

yobert: (also tested and it seems good!)

any1:
Thanks!

PR #131, opened 2022-02-26
Documented wayvnc is now available in Debian as well

gwolf:
Some days ago, the package for wayvnc was accepted into Debian:

https://tracker.debian.org/pkg/wayvnc


any1:
Thanks!

PR #137, opened 2022-05-24
Disable seat completely when using disable-input

tleydxdy:
The first commit is the straight forward patch, the second commit make it so the relevant *_manager aren't even created.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -144,11 +144,6 @@ static void registry_add(void* data, struct wl_registry* registry,
 		return;
 	}
 
-	if (strcmp(interface, wl_shm_interface.name) == 0) {
-		wl_shm = wl_registry_bind(registry, id, &wl_shm_interface, 1);
-		return;
-	}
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -287,25 +288,28 @@ void wayvnc_destroy(struct wayvnc* self)
 
 	wl_shm_destroy(wl_shm);
 
-	if (self->keyboard_backend.virtual_keyboard) {
-		zwp_virtual_keyboard_v1_destroy(
-				self->keyboard_backend.virtual_keyboard);
-		keyboard_destroy(&self->keyboard_backend);
-	}
+	if (!self->disable_input) {
```
any1, line 291: These should all be `NULL` if input is disabled, so there's no need for this.

tleydxdy: fixed, also brown m&ms my fav

any1:
Thanks!

PR #142, opened 2022-07-02
main: check data_control exist before using it

### LINKED ISSUE #141: check data_control exist before using it
image
I got this inconsistent crash, and it turns out it's because my compositor didn't support wlr_data_control but wayvnc happily keep working until it tries to set the selection. maybe it should check that it's inited before doing anything with clipboard.

tleydxdy:
fixes #141


any1:
Thanks!

Issue #80, opened 2020-12-21
Turn off screen capturing when no client is connected


any1:
Wayvnc will continue to capture screens and damage check them while the process is running. Staying idle while nothing is connected would be nicer.

PR #150, opened 2022-08-19
Prevent capture and performance counter when idle

Consolatis:
Fixes
- #80

I have read and understood CONTRIBUTING.md

This PR includes a condition `if (!self->nr_clients)` instead of `if (self->nr_clients == 0)`, will change if you prefer it to be more explicit.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -99,10 +99,14 @@ struct wayvnc {
 	uint32_t n_frames_captured;
 
 	bool disable_input;
+
+	int nr_clients;
+	struct aml_ticker* performance_ticker;
 };
 
 void wayvnc_exit(struct wayvnc* self);
 void on_capture_done(struct screencopy* sc);
+static void on_client_new(struct nvnc_client* client);
 
 #if defined(GIT_VERSION)
 static const char wayvnc_version[] = GIT_VERSION;
@@ -317,6 +321,9 @@ void wayvnc_destroy(struct wayvnc* self)
 
 	if (self->screencopy.manager)
 		zwlr_screencopy_manager_v1_destroy(self->screencopy.manager);
+
+	if (self->performance_ticker)
+		aml_unref(self->performance_ticker);
 
 	wl_registry_destroy(self->registry);
 	wl_display_disconnect(self->display);
@@ -553,6 +560,8 @@ int init_nvnc(struct wayvnc* self, const char* addr, uint16_t port, bool is_unix
 		nvnc_set_key_code_fn(self->nvnc, on_key_code_event);
 	}
 
+	nvnc_set_new_client_fn(self->nvnc, on_client_new);
+
 	nvnc_set_cut_text_receive_fn(self->nvnc, on_client_cut_text);
 
 	struct nvnc_fb* placeholder_fb =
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -770,13 +777,56 @@ static void on_perf_tick(void* obj)
 
 static void start_performance_ticker(struct wayvnc* self)
 {
-	struct aml_ticker* ticker = aml_ticker_new(1000, on_perf_tick, self,
-		NULL);
-	if (!ticker)
+	assert(!self->performance_ticker);
+
+	self->performance_ticker = aml_ticker_new(1000, on_perf_tick, self, NULL);
```
any1, line 782: It's possible to restart a ticker, so it doesn't really need to be reinitialised; just created during init and then started/stopped. Not sure if would make the code simpler or shorter though. You can explore that avenue if you want.

src/main.c
```diff
@@ -770,13 +777,56 @@ static void on_perf_tick(void* obj)
...
+static void on_client_cleanup(struct nvnc_client* client)
+{
+	struct nvnc* nvnc = nvnc_client_get_server(client);
+	struct wayvnc* self = nvnc_get_userdata(nvnc);
+
+	self->nr_clients--;
+	nvnc_log(NVNC_LOG_DEBUG, "Client disconnected, new clients: %d",
+		self->nr_clients);
+	if (!self->nr_clients) {
+		nvnc_log(NVNC_LOG_INFO, "Stopping screencopy");
```
any1, line 808: Maybe say "Stopping screen capturing" instead? "screencopy" is an implementation detail.

src/main.c
```diff
@@ -770,13 +777,56 @@ static void on_perf_tick(void* obj)
...
+		stop_performance_ticker(self);
+	}
+}
+
+static void on_client_new(struct nvnc_client* client)
+{
+	struct nvnc* nvnc = nvnc_client_get_server(client);
+	struct wayvnc* self = nvnc_get_userdata(nvnc);
+
+	if (!self->nr_clients) {
```
any1, line 819: Yeah, this project uses `!` for boolean expressions and `NULL` pointer checks.

src/main.c
```diff
@@ -770,13 +777,56 @@ static void on_perf_tick(void* obj)
...
+		nvnc_log(NVNC_LOG_INFO, "Starting screencopy");
```
any1, line 820: Same as above regarding "screencopy"

any1 (changes requested):

src/main.c
Consolatis: Changed, ticker is created in `main()` and `aml_unref()`ed in `wayvnc_destroy()`.
Additional benefit: the `perfomance_show` boolean in `struct wayvnc` has been removed.

src/main.c
```diff
@@ -1061,13 +1106,12 @@ int main(int argc, char* argv[])
...
-		goto capture_failure;
-
 	if (show_performance)
-		start_performance_ticker(&self);
+		self.performance_ticker = aml_ticker_new(1000, on_perf_tick,
+				&self, NULL);
 
-	wl_display_dispatch(self.display);
+	if (!disable_input)
+		wl_display_dispatch(self.display);
```
Consolatis, line 1114: Force pushed the next iteration:
- Review comments should all be addressed
- Added a check for `disable_input` before `wl_display_dispatch()` in `main()` because it was hanging when starting with `--disable-input` when delaying start of the capture until a client connects

any1: The role of this is to dispatch pending events before blocking. I suppose that without input initialisation, there are no remaining events at that point. A more robust way to do this would be to use `wl_display_dispatch_pending()` since that won't block when there are no events.

Consolatis: Replaced with `wl_display_dispatch_pending()`

any1: Actually, that comment is misleading. It should always have been `wl_display_dispatch_pending()`. The purpose of calling it there is to dispatch any pending events before waiting for new ones. The `aml_poll()` won't detect pending wayland events that have already been read from the socket, so this is there to make sure that they're all handled before blocking. Once inside the main loop, events are handled by `on_wayland_event()` before blocking next time.

Consolatis: Should I remove the comment completely or change into something like
> Make sure all pending events are handled

any1: Yeah, it doesn't need a comment.

Consolatis: deleted the comment

any1:
Thanks!

PR #151, opened 2022-08-20
chore(wrap): use wrap to manager subprojects

Decodetalkers:
If use wrap file , then we do not need to clone by ourself

As the project is using meson, meson can use wrap to manager subprojects, then we need not to use clone anymore

Log: use wrap file to manager subprojects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
+.cache
 build
-subprojects
+#subprojects
 *.swp
 .clang_complete
 .ycm_extra_conf.py
 perf.*
 *.pem
 .vimrc
+/subprojects/**/
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ GitHub [discussion forum](https://github.com/any1/wayvnc/discussions) for the
 project.
 
 ## Installing
-```
+```bash
 # Arch Linux
 pacman -S wayvnc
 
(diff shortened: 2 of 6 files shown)


any1:
This does more than just adding wraps, for which there is already a PR. See #127

Decodetalkers:
> This does more than just adding wraps, for which there is already a PR. See #127 

So it is better to find if there is system dependence first right?

any1:
In the current state of affairs, using the system libraries is the fallback. I.e. if no subprojects are found, the system libraries are used.

The current setup suits me well for development purposes and I don't mind that people have to take a few extra steps to build the project if they're building it themselves. It is very little work compared to the work that I've put in. Also, if you're building it yourself so that you can work on the project, then you'll probably want to set things up in a similar way that I have: with symlinks.

Decodetalkers:
> In the current state of affairs, using the system libraries is the fallback. I.e. if no subprojects are found, the system libraries are used.
> 
> The current setup suits me well for development purposes and I don't mind that people have to take a few extra steps to build the project if they're building it themselves. It is very little work compared to the work that I've put in. Also, if you're building it yourself so that you can work on the project, then you'll probably want to set things up in a similar way that I have: with symlinks.

I think you are right, then I close this pr

PR #154, opened 2022-08-23
main: Prevent capture restart if idle

### LINKED ISSUE #152: "ERROR: ../src/src/main.c: 605: Failed to start capture. Exiting..." when screen resolution changed while wayvnc is running but no client connected
Originally reported as MirServer/mir#2571 - more details like wayland-debug can be seen there.

If you change the resolution of a display while wayvnc is running, but no client connected, the next connection will result in an error:

```plain
ERROR: ../src/src/main.c: 605: Failed to start capture. Exiting...
```

Steps to reproduce:
1. start wayvnc
2. resize the shared display
3. connect a client

Consolatis:
This prevents restarting the capture on dimension change when idle.
Previously the capture was running after the dimensions had changed
which caused starting the capture once a client connects to fail.

Fixes #152

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -614,6 +614,9 @@ void on_output_dimension_change(struct output* output)
 	struct wayvnc* self = output->userdata;
 	assert(self->selected_output == output);
 
+	if (self->nr_clients == 0)
+		return;
+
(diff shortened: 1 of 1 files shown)


any1:
Thanks!

PR #162, opened 2022-10-09
More INFO-level logging

lack:
Adds more output to verbose mode, including:
- Log the listening address and port
- Log removal of seats and outputs
- Add manpage section about multiple outputs
- Log signal handler
- Log output selection and list of outputs

I have (now) read and understood CONTRIBUTING.md and its associated documents.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -534,6 +534,10 @@ int init_nvnc(struct wayvnc* self, const char* addr, uint16_t port, bool is_unix
 		nvnc_log(NVNC_LOG_ERROR, "Failed to bind to address");
 		return -1;
 	}
+	if (is_unix)
+		nvnc_log(NVNC_LOG_INFO, "Listening for connections on %s", addr);
+	else
+		nvnc_log(NVNC_LOG_INFO, "Listening for connections on %s:%d", addr, port);
 
 	self->nvnc_display = nvnc_display_new(0, 0);
 	if (!self->nvnc_display)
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -839,6 +847,17 @@ void parse_keyboard_option(struct wayvnc* self, char* arg)
 	self->kb_layout = arg;
 }
 
+void log_selected_output(struct wayvnc* self)
+{
+	nvnc_log(NVNC_LOG_INFO, "Capturing output %s", self->selected_output->name);
+	struct output* output;
+	wl_list_for_each(output, &self->outputs, link) {
+		nvnc_log(NVNC_LOG_INFO, "%s %s %dx%d+%dx%d",
+				(output->id == self->selected_output->id) ? ">>" : "--",
```
any1, line 856: Consider putting the result of this expression into an intermediate variable so that this can be wrapped within the 80 column limit.

I think it would probably look nicer to use `? "*" : " "`.

any1 (changes requested):
Looks mostly good. Did you read contributing.md?

src/main.c
lack: I did try `? "*" : " "` but since nvnv_log(...) eats all leading spaces, the alignment didn't work.

any1: Ahh. There were some unruly messages coming from ffmpeg. Maybe trimming of leading whitespace was not actually needed. We'll fix it later.

any1:
Thanks!

PR #175, opened 2022-11-08
readme: remove link to compatibility table

emersion:
See https://github.com/any1/neatvnc/commit/36fb346da19527bc2bd645a7570f64ca6fa4956a

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,3 @@ password=p455w0rd
 private_key_file=/path/to/key.pem
 certificate_file=/path/to/cert.pem
 ```
-
-## Compatible Software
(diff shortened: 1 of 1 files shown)


any1:
Thanks!

Lorilandly:
Sorry for posting on a closed thread. But isn't it better to replace the dead link with a link pointing to a working location rather than removing it? 
This link still work [https://github.com/any1/neatvnc/tree/v0.5.4#client-compatibility](url)
Since it's from an old version, I'm not sure the table in this link is accurate tho.

any1:
The table isn't accurate, which is why it was removed.

PR #178, opened 2022-11-12
wayvncctl events

lack:
Adds server->client broadcast events

Read wayvnc.scd and wayvncctl.scd for an overview.

- Ensure fatal errors are sent first
- Add jsonipc event message constructor
- Add server event infrastructure
- Added event help text
- Add generic yaml-like printing of event data
- Add example event-watcher script

### DIFF
--- a/examples/event-watcher
+++ b/examples/event-watcher
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+WAYVNCCTL=${WAYVNCCTL:-wayvncctl}
+
+connection_count_now() {
+    echo "Total clients: $count"
+}
+
+while IFS= read -r EVT; do
+    case "$(jq -r '.method' <<<"$EVT")" in
+        client-*onnected)
+            count=$(jq -r '.params.connection_count' <<<"$EVT")
+            connection_count_now "$count"
+            ;;
+    esac
+done < <("$WAYVNCCTL" --json event-receive)
--- a/src/ctl-client.c
+++ b/src/ctl-client.c
@@ -21,6 +21,7 @@
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <poll.h>
+#include <signal.h>
 #include <jansson.h>
 
 #include "json-ipc.h"
@@ -46,6 +47,8 @@ struct ctl_client {
 
 	char read_buffer[512];
 	size_t read_len;
+	
+	bool wait_for_events;
 
 	int fd;
 };
@@ -197,8 +200,6 @@ static json_t* read_one_object(struct ctl_client* self, int timeout_ms)
 	while (root == NULL) {
 		int n = poll(&pfd, 1, timeout_ms);
 		if (n == -1) {
-			if (errno == EINTR)
-				continue;
 			WARN("Error waiting for a response: %m");
 			break;
 		} else if (n == 0) {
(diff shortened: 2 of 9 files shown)


lack:
Valgrind looks good; no leaks.

Manual testing looks good too.

src/ctl-client.c
```diff
@@ -197,8 +200,6 @@ static json_t* read_one_object(struct ctl_client* self, int timeout_ms)
 	while (root == NULL) {
 		int n = poll(&pfd, 1, timeout_ms);
 		if (n == -1) {
-			if (errno == EINTR)
```
any1, line 200: I suppose you removed this for sake of signal handling? Perhaps it's better to change the condition to `errno == EINTR && self->wait_for_events`

examples/event-watcher
```diff
@@ -0,0 +1,16 @@
...
+}
+
+while IFS= read -r EVT; do
+    case "$(jq -r '.method' <<<"$EVT")" in
+        client-*onnected)
+            count=$(jq -r '.params.connection_count' <<<"$EVT")
+            connection_count_now "$count"
+            ;;
+    esac
+done < <("$WAYVNCCTL" --json event-receive)
```
any1, line 16: Might be convenient to have an option to write the response to an environment variable so you can skip the `read`.

src/ctl-client.c
```diff
@@ -335,13 +373,146 @@ static int ctl_client_print_response(struct ctl_client* self,
...
+	sig_target = self;
+	struct sigaction sa = { 0 };
+	sa.sa_handler = stop_loop;
+	sigaction(SIGINT, &sa, NULL);
+	sigaction(SIGTERM, &sa, NULL);
+}
+
+static void print_indent(int level)
+{
+	for(int i = 0; i < level; ++i)
```
any1, line 400: Nit: style

src/ctl-server.c
```diff
@@ -303,14 +363,28 @@ static json_t* client_next_object(struct ctl_client* self, struct cmd_response**
 	return root;
 }
 
-static struct cmd_response* generate_help_object(const char* cmd)
+static struct cmd_info* find_info(const char* id, struct cmd_info (*list)[],
+		size_t len)
+{
+	for(int i = 0; i < len; ++i) {
```
any1, line 369: Nit: style

wayvncctl.scd
```diff
@@ -42,6 +42,14 @@ a list of the available commands.
 Running *wayvncctl command-name --help* returns a description of the server-side
 command and its available parameters.
 
+# ASYNCHRONOUS EVENTS
+
+While *wayvncctl* normally terminates after sending one request and receiving
+the corresponding reply, the *event-receive* command acts differently. Instead
+of exiting immediately, *wayvncctl* waits for any events fr the server, printing
```
any1, line 49: "events fr server"?

wayvncctl.scd
```diff
@@ -42,6 +42,14 @@ a list of the available commands.
...
+each to stdout as they arrive. This mode of operation will block until either
+it receives a signal to terminate, or until the wayvnc server terminates.
+
```
any1, line 52: It would be good to state how events are delimited. I.e. do they use newline or something else?

src/ctl-server.c
```diff
@@ -377,8 +456,19 @@ static void client_set_aml_event_mask(struct ctl_client* self)
...
+		result = json_array_insert(self->response_queue, 0, message);
+	else
+		result = json_array_append(self->response_queue, message);
+	client_set_aml_event_mask(self);
+	return result;
+}
+
 static int client_enqueue_jsonipc(struct ctl_client* self,
-		struct jsonipc_response* resp)
+		struct jsonipc_response* resp, bool at_head)
```
any1, line 471: Nit: It's better to use enums instead of boolean arguments because they make the code easier to read at a glance.

wayvncctl.scd
lack: Phone keyboard 🤦

examples/event-watcher
lack: Trying to think of how we can do that.

One idea I had was to introduce a "script mode" where a user could run `wayvncctl --event-script=/path/to/executable` and then we would run that executable once for every event, with some reasonable calling convention that puts the event details in the env and/or cmdline.  But I think I'd want to do that in a separate PR.

any1:
Thanks!

PR #183, opened 2022-11-13
Cleanup stale unix socket, iff inactive

### LINKED ISSUE #180: Warning: ../src/ctl-server.c: 633: Failed to bind unix socket: Address already in use
When trying to start wayvnc I get this error:
`Warning: ../src/ctl-server.c: 633: Failed to bind unix socket: Address already in use`
and wayvnc does not start.
I think this is because my previous instance of wayvnc did not shut down cleanly and left this socket in the file system:
```
$ ls -l $XDG_RUNTIME_DIR/wayvncctl 
srwxrwxr-x 1 hch hch 0 nov 11 15:58 /run/user/1000/wayvncctl=

```
This should have been removed in ctl_server_stop but that was never run and now wayvnc can't start. I can remove it manually but I do think wayvnc should always remove this before starting up.

lack:
Cleans up a pre-existing control socket but only if we can't connect to it.

Fixes #180

### DIFF
--- a/src/ctl-server.c
+++ b/src/ctl-server.c
@@ -21,6 +21,7 @@
 #include <errno.h>
 #include <sys/socket.h>
 #include <sys/un.h>
+#include <sys/stat.h>
 #include <netdb.h>
 #include <neatvnc.h>
 #include <aml.h>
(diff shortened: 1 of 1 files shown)


any1:
Very succinct commit message. Thanks!

PR #185, opened 2022-11-16
Add --wait and --reconnect to wayvncctl

lack:
- Fix wayvncctl usage message typo
- Add wayvncctl --wait option
- Add '--reconnect' mode for the wayvncctl event loop

It occurred to me that when running wayvntctl in a loop to listen for events, it would be useful to be able to persist the loop even when wayvnc isn't running.  So I've added:

`wayvncctl --wait` that that works for the event loop as well as any other command.  As it starts up, if the control socket doesn't exist, it will wait indefinitely, waiting for the socket to be created.

`wayvncctl --reconnect` only works for the event-recieve loop.  Basically if wayvnc is terminated, the event loop normally terminates too.  But with this flag set, it will go into a loop that waits for the control socket to be re-created, and when it comes back it re-registers for events and continues receiving events.


src/ctl-client.c
```diff
@@ -39,11 +41,12 @@ static bool do_debug = false;
 	if (do_debug) \
 		fprintf(stderr, "[%s:%d] " fmt "\n", __FILE__, __LINE__, ##__VA_ARGS__);
 
-#define FAILED_TO(action) \
-	WARN("Failed to " action ": %m");
+#define FAILED_TO(action, ...) \
```
any1, line 44: Maybe this `FAILED_TO` macro isn't very useful to begin with. Maybe it's better to just write things out...

One problem that I see with it is that it interferes with grepping for error messages.

src/ctl-client.c
```diff
@@ -65,37 +68,76 @@ struct ctl_client* ctl_client_new(const char* socket_path, void* userdata)
...
 socket_failure:
 	free(new);
 	return NULL;
 }
 
+int ctl_client_connect(struct ctl_client* self, int timeout)
+{
+	// TODO: Support arbitrary timeouts?
+	assert(timeout == 0 || timeout == -1);
+	// TODO: Use inotify instead of polling stat()
```
any1, line 91: I don't think that `inotify` is portable. Polling is fine here.

src/ctl-client.c
```diff
@@ -65,37 +68,76 @@ struct ctl_client* ctl_client_new(const char* socket_path, void* userdata)
...
 	return new;
 
-connect_failure:
-	close(new->fd);
 socket_failure:
 	free(new);
 	return NULL;
 }
 
+int ctl_client_connect(struct ctl_client* self, int timeout)
```
any1, line 87: It would be good to split this into smaller functions to aide readability.

src/ctl-client.c
```diff
@@ -65,37 +68,76 @@ struct ctl_client* ctl_client_new(const char* socket_path, void* userdata)
...
+{
+	// TODO: Support arbitrary timeouts?
+	assert(timeout == 0 || timeout == -1);
+	// TODO: Use inotify instead of polling stat()
+	bool needs_log = true;
+	struct stat sb;
+	while (stat(self->addr.sun_path, &sb) != 0) {
```
any1, line 94: For example: this block here could be a function called something like `ctl_client__wait_until_socket_exists`

any1 (commented):
LGTM. Suggested changes are optional.

src/ctl-client.c
lack: Perhaps. The alternative seems pretty copy-pastey, though...

src/ctl-client.c
lack: Correct. Linux only. Thoug  it could be ifdef'd.  I'll play around with it maybe, but not in this PR.

src/ctl-client.c
lack: I think that's a very good idea.  Will do.

src/ctl-client.c
any1: I prefer copy-pastey to a macro that only adds 14 static characters to the string. :)

lack:
@any1 : Okay, ready for final review!

any1:
Thanks!

src/ctl-client.c
any1: I would not recommend doing ifdefs for that. This is good enough and adding platform dependent code just means that there's more code to maintain. The fallback path would also be untested most of the time because we probably won't bother disabling `inotify` just to test that one code-path.

If you want to add `inotify` as an academic exercise, go ahead, but I don't think that I will merge such a change. :)

PR #186, opened 2022-11-19
Introduce wayvncctl startup amd shutdown events

lack:
Especially useful when using `--wait` or `--reconnect` mode, wayvncctl will
generate 2 additional events not documented in *wayvnc(1)*:

*wayvnc-startup*
	Sent when a successful wayvnc control connection is established and
	event registration has succeeded, both upon initial startup and on
	subsequent registrations with `--reconnect`.

	No paramerers.

*wayvnc-shutdown*
	Sent when the wayvnc control connection is dropped, usually due to
	wayvnc exiting.

	No paramerers.

### DIFF
--- a/examples/event-watcher
+++ b/examples/event-watcher
@@ -3,7 +3,7 @@
 WAYVNCCTL=${WAYVNCCTL:-wayvncctl}
 
 connection_count_now() {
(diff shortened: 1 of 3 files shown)


any1 (approved):
LGTM

PR #189, opened 2022-11-22
Add a PR-checker action that makes sure CONTRIBUTING.md was read

lack:

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -1,0 +1,30 @@
+name: CONTRIBUTING.md compliance
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+
+jobs:
+  pr_description:
+    runs-on: ubuntu-latest
+    steps:
+    - name: PR description contains appropriate notice text
+      run: |-
+        tr -s '[:space:]' ' ' \
+          <<< '${{ github.event.pull_request.body }}' \
+        | \
+        grep -qF \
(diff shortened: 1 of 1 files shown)


Consolatis:
My 2 cents:

Personally I feel a bit alienated by things like this (including the original `I have read and understood CONTRIBUTING.md and its associated documents.` requirement, I think I even used something like this before: I have read and understood CONTRIBUTING.md ~~and its associated documents~~.).
I don't like to get told to read random external coding guidelines for stuff that I might not even use within a PR.
It's my free time I invest after all. If I got something wrong I'm very happy to fix the PR, if I am a regular contributor I will also read all the external stuff. Pointing to Contributing.md when opening a PR is totally valid but having to write that sentence goes a bit too far IMHO.

We started to [include](https://github.com/labwc/labwc/tree/master/scripts) a slightly modified checkpatch.pl in our tree and are thinking about letting the CI run it on PRs. IMO this is a much friendlier way for contributors to ensure some specific code style while also reducing the time required for reviewing.

But again, just my 2 cents.

any1:
This "canary" text in the contributing guide there so that I can get an idea about how much work is going to be required from me to review the PR. If it's a large PR, I might take and cursory look over the changes and if I see failures to adhere to the guide in both the changes and the PR summary, I will request that the submitter read the guide. Otherwise, I will point out the infractions.

I agree that it would be better to have the entire guide in a single document, but that requires time and effort, which I've elected to spend elsewhere to this day.

A linter is perhaps ideal, but making it work properly also takes time and effort.

As for this proposed change, it is perhaps better to shelve it for now as I do not wish to enforce this rule for all PRs.

lack:
Shelving!

PR #191, opened 2022-11-22
Fix segfault if running wayvncctl with no args

lack:


any1 (approved):
LGTM

PR #193, opened 2022-11-25
Start capture on output switch iff clients are connected

lack:


any1:
Thanks!

PR #195, opened 2022-11-26
Add a brief timeout between a capture failure and the retry

### LINKED ISSUE #194: High CPU usage when wayvnc is capturing an output that is "DPMS off"
How to replicate:
- Start wayvnc
- Measure CPU usage (minimal)
- turn screens off with `swaymsg output '*' DPMS off`
- Connect a VNC client
- Measure CPU usage. I see sway and wayvnc using ~50% of a single core, each.
- turn screens on with `swaymsg output '*' DPMS on` or disconnect the client
- Measure CPU usage (much lower)

lack:
Ensures we don't spin eating CPU if capture is continually fsiling due
to DPMS (for example) by only trying to restart screen capture every
100ms.

Fixes #194

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -104,6 +104,8 @@ struct wayvnc {
 	int nr_clients;
 	struct aml_ticker* performance_ticker;
 
+	struct aml_timer* capture_retry_timer;
+
 	struct ctl* ctl;
 };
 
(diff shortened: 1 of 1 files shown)


any1:
Thanks!

PR #197, opened 2022-11-30
DPMS Support

### LINKED ISSUE #114: can't connect to wayvnc when dpms off
My laptop sway's idle configuration:
 exec swayidle -w \
          timeout 300 'swaylock -f -c 000000' \
          timeout 600 'swaymsg "output * dpms off"' resume 'swaymsg "output * dpms on"' \
          before-sleep 'swaylock -f -c 000000'

1. wait 600s or run swaymsg "output * dpms off"' from ssh
2. start wayvnc
3. use another pc vnc-client to connect to laptop's wayvnc

vnc-client can't connect to wayvnc.

lack:
- Add DPMS state to the output object
- Don't attempt capture if an output is DPMS off
- Turn DPMS on when starting capture
- Blank the capture screen when DPMS is turned off

Fixes #114

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -39,6 +39,7 @@
 #include "wlr-virtual-pointer-unstable-v1.h"
 #include "virtual-keyboard-unstable-v1.h"
 #include "xdg-output-unstable-v1.h"
+#include "wlr-output-power-management-unstable-v1.h"
 #include "linux-dmabuf-unstable-v1.h"
 #include "screencopy.h"
 #include "data-control.h"
@@ -78,6 +79,7 @@ struct wayvnc {
 	struct cfg cfg;
 
 	struct zxdg_output_manager_v1* xdg_output_manager;
+	struct zwlr_output_power_manager_v1* wlr_output_power_manager;
 	struct zwp_virtual_keyboard_manager_v1* keyboard_manager;
 	struct zwlr_virtual_pointer_manager_v1* pointer_manager;
 	struct zwlr_data_control_manager_v1* data_control_manager;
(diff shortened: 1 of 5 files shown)


any1:
Pretty good. We should probably make this protocol optional, since it is vendor specific and IIRC not all of it is implemented inside wlroots. Some of it needs to be implemented in the compositor, so they might not have implemented it.

lack:
> Pretty good. We should probably make this protocol optional, since it is vendor specific and IIRC not all of it is implemented inside wlroots. Some of it needs to be implemented in the compositor, so they might not have implemented it.

Good point.  Will do!  I think it's about 80% optional right now ;)

lack:
@any1 I think that makes it properly optional, though I'm still unclear on how these wayland extensions work.

But as far as I can figure, if the registry doesn't see the interface name, the `wayvnc->wlr_output_power_manager` will remain null, and the DPMS-based logic shouldn't take effect.

lack:
Tested by temporarily commenting out the registry detection logic; Looked good; didn't crash :)

any1:
https://github.com/swaywm/sway/pull/7083

Perhaps we ought to speak of "power" instead of "dpms"?

src/main.c
```diff
@@ -1107,11 +1167,14 @@ void log_selected_output(struct wayvnc* self)
...
 	wl_list_for_each(output, &self->outputs, link) {
 		bool this_output = (output->id == self->selected_output->id);
-		nvnc_log(NVNC_LOG_INFO, "%s %s %dx%d+%dx%d",
+		nvnc_log(NVNC_LOG_INFO, "%s %s %dx%d+%dx%d DPMS:%s",
 				this_output ? ">>" : "--",
 				output->description,
 				output->width, output->height,
-				output->x, output->y);
+				output->x, output->y,
+				(output->dpms == OUTPUT_DPMS_ON) ? "ON" :
```
any1, line 1175: Nit: I would make a small utility function called something like `output_power_state_name`.

lack:
@any1 Okay!  Removed any mention of DPMS from code and log, and added the to-string helper.

any1:
Thanks!

PR #199, opened 2022-12-07
Add output power state to get_outputs ipc command

lack:
I forgot to plumb this through when I did the initial state on the
output power state.


any1 (approved):
LGTM

Issue #45, opened 2020-07-01
SEGV whenever other outputs change 


colemickens:
1. I was VNC'd into my laptop running Sway. At this time, the KVM was not on the laptop, so the only output was HEADLESS-1.
2. The KVM switched back to the laptop, enabling the other output.

At this point my VNC session crashed. Sway kept running and I was able to just restart wayvnc, but it would be nice if it didn't crash of course.

```
           PID: 28857 (wayvnc)
           UID: 1000 (cole)
           GID: 1000 (cole)
        Signal: 11 (SEGV)
     Timestamp: Wed 2020-07-01 16:00:06 PDT (10min ago)
  Command Line: wayvnc -o HEADLESS-1 0.0.0.0
    Executable: /nix/store/ravgkjx5c1bssi4svnnbh05cz1p05508-wayvnc-4f1bca55c3bf19a58e9d892a5141fdfc5159bf01/bin/wayvnc
 Control Group: /user.slice/user-1000.slice/session-8.scope
          Unit: session-8.scope
         Slice: user-1000.slice
       Session: 8
     Owner UID: 1000 (cole)
       Boot ID: 7afde54d612c4eaebc38d2961d43f7ed
    Machine ID: befa945b3fb44f73a9671674ee23e039
      Hostname: xeep
       Storage: /var/lib/systemd/coredump/core.wayvnc.1000.7afde54d612c4eaebc38d2961d43f7ed.28857.1593644406000000000000.lz4
       Message: Process 28857 (wayvnc) of user 1000 dumped core.

Reading symbols from /nix/store/ravgkjx5c1bssi4svnnbh05cz1p05508-wayvnc-4f1bca55c3bf19a58e9d892a5141fdfc5159bf01/bin/wayvnc...
(No debugging symbols found in /nix/store/ravgkjx5c1bssi4svnnbh05cz1p05508-wayvnc-4f1bca55c3bf19a58e9d892a5141fdfc5159bf01/bin/wayvnc)
[New LWP 28857]
[New LWP 28864]
[New LWP 28858]
[New LWP 28859]
[New LWP 28860]
[New LWP 28861]
[New LWP 28867]
[New LWP 28869]
[New LWP 28865]
[New LWP 28866]
[New LWP 28868]
[New LWP 28863]
[New LWP 28870]
[New LWP 28862]

warning: Unexpected size of section `.reg-xstate/28857' in core file.
[Thread debugging using libthread_db enabled]
[...]

PR #201, opened 2022-12-18
Fix segfault after hot-plugging outputs

lack:
When new outputs appear, either because they were just created or
because they have been disabled/reenabled, ensure we always set up the
right xdg_output and wlr_output_power listeners.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -78,8 +78,6 @@ struct wayvnc {
 	struct wl_list seats;
 	struct cfg cfg;
 
-	struct zxdg_output_manager_v1* xdg_output_manager;
-	struct zwlr_output_power_manager_v1* wlr_output_power_manager;
 	struct zwp_virtual_keyboard_manager_v1* keyboard_manager;
 	struct zwlr_virtual_pointer_manager_v1* pointer_manager;
 	struct zwlr_data_control_manager_v1* data_control_manager;
@@ -134,6 +132,8 @@ static void client_init_data_control(struct wayvnc_client* self);
 struct wl_shm* wl_shm = NULL;
 struct zwp_linux_dmabuf_v1* zwp_linux_dmabuf = NULL;
 struct gbm_device* gbm_device = NULL;
+struct zxdg_output_manager_v1* xdg_output_manager = NULL;
+struct zwlr_output_power_manager_v1* wlr_output_power_manager = NULL;
 
 static bool registry_add_input(void* data, struct wl_registry* registry,
 			 uint32_t id, const char* interface,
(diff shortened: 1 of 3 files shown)


lack:
Fixes #45

src/main.c
```diff
@@ -413,9 +414,6 @@ static int init_wayland(struct wayvnc* self)
 		goto failure;
 	}
 
-	wl_display_dispatch(self->display);
-	wl_display_roundtrip(self->display);
-
```
lack, line 418: @any1 : I wanted to double-check on this part, as I'm still a Wayland newbie.

As far as I can see this was only needed because of the events added by `init-outputs()`, and by moving the output listeners registrationd to the registry callback, those events get handled in the earlier dispatch/roundtrip now.

Plus, if I leave this in wayvnc hangs here now.

any1: Yes, that seems about right. Frankly, I think the order of those is wrong here anyway. It should be `roundtrip(); dispatch()`. The round trip flushes and waits for an answer and the dispatch dispatches callbacks for the answer.

any1 (commented):
I think that it would probably be best to just move the responsibility of initialising `xdg_output` and `wlr_output_power` into `output.c`.

The managers just need to be made global variables so that they can be accessed from `output.c`.  This is what I've done with `wl_shm` and `zwp_linux_dmabuf`.

src/main.c
lack: I'll flip it and see what happens.

any1: It's done in this order in other places too. That should be fixed, but that's (arguably) a different issue.

lack: The actual behavior seems opposite... If I reverse the of the calls, I don't get some of the events until the main event loop.

lack:
I think this is better now.  More logic in output.c, less in main.c

lack:
@any1 Ready for re-review

src/output.c
```diff
@@ -199,24 +203,6 @@ void output_list_destroy(struct wl_list* list)
 	}
 }
 
-struct output* output_new(struct wl_output* wl_output, uint32_t id)
```
lack, line 202: Moves down to call static methods

src/output.c
```diff
@@ -310,9 +302,12 @@ static const struct zwlr_output_power_v1_listener wlr_output_power_listener = {
 	.failed = output_power_failed,
 };
 
-void output_set_wlr_output_power(struct output* self,
-                           struct zwlr_output_power_v1* wlr_output_power)
+static void output_setup_wlr_output_power_manager(struct output* self)
 {
+	struct zwlr_output_power_v1* wlr_output_power =
```
any1, line 307: Don't you need to check here if `self->wlr_output_power` has already been set?

lack: I think I missed that in my commit somehow have it locally!

any1 (approved):
LGTM

PR #202, opened 2022-12-18
Custom option parser

any1:
This implements a custom option parser.
The individual commit history is not great, but I've kept it for the sake of review. The commits will be squashed before merging for the sake of bisectability.

The new parser allows us to define positional options and subcommands in the "option schema". This helps with consistency and facilitates abstractions.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -20,7 +20,6 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <unistd.h>
-#include <getopt.h>
 #include <assert.h>
 #include <inttypes.h>
 #include <errno.h>
@@ -1059,7 +1058,7 @@ static void on_nvnc_client_new(struct nvnc_client* client)
 			self->nr_clients);
 }
 
-void parse_keyboard_option(struct wayvnc* self, char* arg)
+void parse_keyboard_option(struct wayvnc* self, const char* arg)
 {
 	// Find optional variant, separated by -
 	char* index = strchr(arg, '-');
@@ -1242,6 +1241,7 @@ int main(int argc, char* argv[])
 	const char* output_name = NULL;
 	const char* seat_name = NULL;
 	const char* socket_path = NULL;
+	const char* keyboard_options = NULL;
 
 	bool overlay_cursor = false;
 	bool show_performance = false;
@@ -1253,6 +1253,8 @@ int main(int argc, char* argv[])
 	int log_level = NVNC_LOG_WARNING;
 
 	static const struct wv_option opts[] = {
+		{ .positional = "address" },
+		{ .positional = "port" },
 		{ 'C', "config", "<path>",
 		  "Select a config file." },
 		{ 'g', "gpu", NULL,
@@ -1268,7 +1270,8 @@ int main(int argc, char* argv[])
 		{ 'r', "render-cursor", NULL,
 		  "Enable overlay cursor rendering." },
 		{ 'f', "max-fps", "<fps>",
-		  "Set rate limit (default 30)." },
+		  "Set rate limit (default 30).",
+		  .default_ = "30" },
 		{ 'p', "performance", NULL,
 		  "Show performance counters." },
 		{ 'u', "unix-socket", NULL,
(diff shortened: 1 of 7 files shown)


lack:
For subcommands, is there a way to register a list of valid subcommands, and provide a second level of command-specific options for each one?

any1:
A subcommand ends the processing of arguments. The idea is to run the option parser again for each subcommand.

src/wayvncctl.c
```diff
@@ -85,46 +86,35 @@ int main(int argc, char* argv[])
...
+	flags |= option_parser_get_value(&option_parser, "json")
+		? CTL_CLIENT_PRINT_JSON : 0;
+	verbose = !!option_parser_get_value(&option_parser, "verbose");
+
+	// No command; nothing to do...
+	if (!option_parser_get_value(&option_parser, "command"))
+		return 0;
+
+	argc -= option_parser.endpos;
+	argv += option_parser.endpos;
```
lack, line 117: Would it be cleaner to just have `option_parser.argc` and `option_parser.argv` that we could pass on to the next level?

lack (approved):
Looking good! A few minor optional suggestions in comments

src/main.c
```diff
@@ -1292,84 +1295,52 @@ int main(int argc, char* argv[])
 		{ 'v', "verbose", NULL,
 		  "Be more verbose. Same as setting --log-level=info" },
 		{ 'L', "log-level", "<level>",
-		  "Set log level. The levels are: error, warning, info, debug trace and quiet." },
+		  "Set log level. The levels are: error, warning, info, debug trace and quiet.",
```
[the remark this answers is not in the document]
any1: That could be added as a feature to the option parser, but it's probably simpler and more flexible to just parse and check the values afterwards.

This is the question of framework vs. toolkit. The option parser is turning into a bit of a framework, but I don't think that we should take it all the way and give it all the bells and whistles.

The goals for the option parser interface that is currently in master were the following:
 * Consolidate usage text with option definitions. This keeps them in the same place so they're easier to keep in sync.
 * Make it possible to generate option parser schemas from `cmd_info`, allowing us to also use the option parser to parse subcommand options and to generate usage text for them.

The problem with the previous implementation was that it lacks a way to describe and parse positional arguments. That's pretty much the whole reason for creating a custom parser like this: so that we can have schema based positional arguments.

src/main.c
```diff
@@ -1280,7 +1282,8 @@ int main(int argc, char* argv[])
 		{ 'r', "render-cursor", NULL,
 		  "Enable overlay cursor rendering." },
 		{ 'f', "max-fps", "<fps>",
-		  "Set rate limit (default 30)." },
+		  "Set rate limit (default 30).",
```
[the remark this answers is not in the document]
any1: Only if it's useful for `waynvcctl` subcommand options.

src/wayvncctl.c
any1: Don't know. There are many things about this interface that could be improved. I think we'll figure out ways to improve the ergonomics of it along the way.

src/main.c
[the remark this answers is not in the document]
lack: Probably not yet.

src/option-parser.c
```diff
@@ -171,7 +149,223 @@ void option_parser_print_options(struct option_parser* self, FILE* stream)
...
+
+		current_pos += 1;
+	}
+	return NULL;
+}
+
+static int append_value(struct option_parser* self,
+		const struct wv_option* option, const char* value)
+{
+	if ((size_t)self->n_values >= ARRAY_SIZE(self->values)) {
```
lack, line 193: Would it be cleaner/more correct if n_values, n_opts, and other related position values in the structure were `size_t` instead of `int`?

lack (approved):
LGTM with a minor suggestion

lack:
@any1 Justcpushed a commit to this branch implementing my suggestion. Feel free to squash it and merge, or delete it and merge, whichever you like!

src/option-parser.c
any1: I like `int` for the following reasons:
 * It's easier to type
 * `argc` is `int`
 * It's signed, so intermediate steps in arithmetic may yield negative values without adverse effects.

src/main.c
```diff
@@ -1280,84 +1283,52 @@ int main(int argc, char* argv[])
...
-		case 'V':
-			return show_version();
-		case 'h':
-			return wayvnc_usage(&option_parser, stdout, 0);
-		default:
-			return wayvnc_usage(&option_parser, stderr, 1);
-		}
+	option_parser_init(&option_parser, opts);
+
+	if (option_parser_parse(&option_parser, argc, (const char**)argv) < 0)
```
any1, line 1296: This cast to `const char` should be eliminated. It's a hack that I put in there because I was lazy.

lack: I think the cast is the right thing to do. C can't implicitly cast this, so I think the explicit cast is the right choice.

Reference: https://c-faq.com/ansi/constmismatch.html

How would you propose to fix it?  One option is to change `main()` so argv is const. The other is to just change the signature of `option_parser_parse` to accept a non-const argv...

I vote to just live with the cast.

any1: Huh, I actually didn't know about this limitation. Anecdotally, `char* const*` works, but `const char* const*` does not. Anyway, this should really be `const char* const*`.

PR #203, opened 2022-12-20
Prevent race condition with newly detected outputs

lack:
This change delays the addition of a new output to the wayvnc->outputs
list until all initial registration events settle, preventing any chance
of using an incompletely-initialized output.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -107,6 +107,7 @@ struct wayvnc {
 	struct aml_timer* capture_retry_timer;
 
 	struct ctl* ctl;
+	bool is_initializing;
 };
 
 struct wayvnc_client {
@@ -201,6 +202,10 @@ static void registry_add(void* data, struct wl_registry* registry,
 			return;
 
 		wl_list_insert(&self->outputs, &output->link);
+		if (!self->is_initializing) {
+			wl_display_dispatch(self->display);
+			wl_display_roundtrip(self->display);
+		}
 		return;
 	}
 
@@ -364,6 +369,7 @@ void wayvnc_destroy(struct wayvnc* self)
 
 static int init_wayland(struct wayvnc* self)
 {
+	self->is_initializing = true;
 	static const struct wl_registry_listener registry_listener = {
 		.global = registry_add,
 		.global_remove = registry_remove,
@@ -384,6 +390,7 @@ static int init_wayland(struct wayvnc* self)
 
 	wl_display_dispatch(self->display);
 	wl_display_roundtrip(self->display);
+	self->is_initializing = false;
 
 	if (!self->pointer_manager && !self->disable_input) {
 		nvnc_log(NVNC_LOG_ERROR, "Virtual Pointer protocol not supported by compositor.");


lack:
@any1 I would like your eyes on this one. Basically I'm deferring adding the output objects to the `wayvnc->outputs` list until all registered listeners each get their first set of events.

This way the only objects in the list are truly really and fully initialized, eliminating a race that exists if I have a script that creates a headless output and then tries immediately to start capturing it.

any1:
It's unlikely, but not putting the output into the list immediately creates at least one other race condition. I.e. the output could go away again before the "ready" event and then the object will leak, or worse.

Maybe it's better to just check if the output is complete when traversing through the list?

any1:
I feel like having an intermediate list adds more to the overall complexity than just skipping incomplete outputs in `output_cycle`, `output_first`, etc...

I haven't had time to give this a really close look though.

Another thing that we can do is to force a roundtrip, when a new output is added. This would guarantee that everything is initialised when `output_new` returns. It's very heavy-handed and might cause a slight stutter anytime that a new output is added, but it keeps things simple and gets the job done.

lack:
Perhaps.  I'll try an alternative implementation that just skips by comparison.

But regardless I think I want to hide the wl_list head behind an interface to discourage using wl_list_for_each directly in main.c

lack:
Not sure how I feel about this.  I may experiment with your "round-trip on output_new" idea and see how that affects the connection.

any1:
Isn't it enough to do something like this?

```
diff --git a/include/output.h b/include/output.h
index 5e9a6a6..a500082 100644
--- a/include/output.h
+++ b/include/output.h
@@ -74,6 +74,8 @@ struct output* output_find_by_id(struct wl_list* list, uint32_t id);
 struct output* output_find_by_name(struct wl_list* list, const char* name);
 struct output* output_first(struct wl_list* list);
 
+bool output_is_complete(struct output* self);
+
 enum output_cycle_direction {
 	OUTPUT_CYCLE_FORWARD,
 	OUTPUT_CYCLE_REVERSE,
diff --git a/src/main.c b/src/main.c
index b719f3d..afebace 100644
--- a/src/main.c
+++ b/src/main.c
@@ -123,7 +123,6 @@ void wayvnc_exit(struct wayvnc* self);
 void on_capture_done(struct screencopy* sc);
 static void on_nvnc_client_new(struct nvnc_client* client);
 void switch_to_output(struct wayvnc*, struct output*);
-void switch_to_next_output(struct wayvnc*);
 void switch_to_prev_output(struct wayvnc*);
 static void client_init_pointer(struct wayvnc_client* self);
 static void client_init_keyboard(struct wayvnc_client* self);
@@ -449,8 +448,9 @@ struct cmd_response* on_output_cycle(struct ctl* ctl, enum output_cycle_directio
 	struct wayvnc* self = ctl_server_userdata(ctl);
 	nvnc_log(NVNC_LOG_INFO, "ctl command: Rotating to %s output",
 			direction == OUTPUT_CYCLE_FORWARD ? "next" : "previous");
-	struct output* next = output_cycle(&self->outputs,
-			self->selected_output, direction);
+	struct output* next;
+	do next = output_cycle(&self->outputs, self->selected_output, direction);
+	while (next && !output_is_complete(next));
 	switch_to_output(self, next);
 	return cmd_ok();
 }
@@ -463,7 +463,7 @@ struct cmd_response* on_output_switch(struct ctl* ctl,
 	if (!output_name || output_name[0] == '\0')
 		return cmd_failed("Output name is required");
 	struct output* output = output_find_by_name(&self->outputs, output_name);
-	if (!output) {
+	if (!output || !output_is_complete(output)) {
 		return cmd_failed("No such output \"%s\"", output_name);
 	}
 	switch_to_output(self, output);
@@ -1185,20 +1185,13 @@ void switch_to_output(struct wayvnc* self, struct output* output)
 		wayvnc_start_capture_immediate(self);
 }
 
-void switch_to_next_output(struct wayvnc* self)
-{
-	nvnc_log(NVNC_LOG_INFO, "Rotating to next output");
-	struct output* next = output_cycle(&self->outputs,
-			self->selected_output, OUTPUT_CYCLE_FORWARD);
-
-	switch_to_output(self, next);
-}
-
 void switch_to_prev_output(struct wayvnc* self)
 {
 	nvnc_log(NVNC_LOG_INFO, "Rotating to previous output");
-	struct output* prev = output_cycle(&self->outputs,
-			self->selected_output, OUTPUT_CYCLE_REVERSE);
+	struct output* prev;
+	do prev = output_cycle(&self->outputs, self->selected_output,
+			OUTPUT_CYCLE_REVERSE);
+	while (prev && !output_is_complete(prev));
 	switch_to_output(self, prev);
 }
 
diff --git a/src/output.c b/src/output.c
index 7773c51..d3ce10d 100644
--- a/src/output.c
+++ b/src/output.c
@@ -410,3 +410,9 @@ struct output* output_new(struct wl_output* wl_output, uint32_t id)
 
 	return output;
 }
+
+bool output_is_complete(struct output* self)
+{
+	return self->xdg_output &&
+		(!wlr_output_power_manager || self->wlr_output_power);
+}
```

By the way, when looking over this, I noticed that it seems that we don't really handle the case where all outputs go away. Shouldn't we handle that somehow? E.g. by switching to the first output that becomes available?

lack:
Nearly, but we should also hide incomplete outputs from the other places that use the list, like the `get-outputs` IPC.

This latest version just forces a round-trip whenever a new output is registered. In my testing I didn't notice any disruption of the VNC session.  And it's so much simpler.

any1 (approved):
I think this will do for now. The simplicity is quite appealing.

PR #210, opened 2023-01-02
Convert wayvncctl subcommands to use option-parser

lack:

### DIFF
--- a/src/ctl-client.c
+++ b/src/ctl-client.c
@@ -32,6 +32,7 @@
 #include "ctl-server.h"
 #include "strlcpy.h"
 #include "util.h"
+#include "option-parser.h"
 
 #define LOG(level, fmt, ...) \
 	fprintf(stderr, "[%s:%d] <" level "> " fmt "\n", __FILE__, __LINE__, \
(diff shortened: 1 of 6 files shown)


test/option-parser-test.c
```diff
@@ -129,6 +159,9 @@ static int test_stop(void)
 
 	ASSERT_TRUE(option_parser_get_value(&parser, "a"));
 	ASSERT_FALSE(option_parser_get_value(&parser, "b"));
+	ASSERT_INT_EQ(2, parser.remaining_argc);
+	ASSERT_STR_EQ("--", parser.remaining_argv[0]);
+	ASSERT_STR_EQ("-b", parser.remaining_argv[1]);
```
lack, line 164: @any1 : Question about API flavor... 

- Should this skip the `--` and just return the remaining args (ie argv[0] would be "-b") so it's more like how extra positional arguments are portrayed?

- Or should `--` be argv[0] (as it is now) so it's more like how subcommands work?

any1: It's more consistent like it is, but it also doesn't really make a lot of sense because `--` is a dud. I would skip it.

src/ctl-client.c
```diff
@@ -731,17 +711,62 @@ static int ctl_client_print_help(struct ctl_client* self,
...
+		printf("Invalid command");
+		return -1;
+	}
+
+	size_t param_count = 0;
+	while (info->params[param_count].name != NULL)
+		param_count++;
+
+	// Add 2: one for --help and one to null-terminate the list
+	struct wv_option* options = calloc(param_count + 2,
```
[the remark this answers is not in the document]
lack: :facepalm: yup.

PR #216, opened 2023-01-06
Remove the help command from wayvncctl

### LINKED ISSUE #213: Remove `wayvncctl help`
As discussed in IRC, there's no need for this way of invoking the help any more, and as evidenced in #211, supporting more than one way to show help text is just asking for weird bugs.

We should remove the `help` command from wayvncctl entirely.

- For command help, `wayvncctl command-name --help` works great.
- For event help, this should be part of the `wayvncctl event-receive --help` output.
- Need to remember to update documentation to match, too.

lack:
- Remove 'help' command and clean up help output
- Add help text for wayvncctl-only events
- Refactor option-parser table printing code for reuse
- Add command and event details to help output

Fixes #213

### DIFF
--- a/src/ctl-client.c
+++ b/src/ctl-client.c
@@ -25,6 +25,7 @@
 #include <signal.h>
 #include <assert.h>
 #include <jansson.h>
+#include <sys/param.h>
 
 #include "json-ipc.h"
 #include "ctl-client.h"
@@ -33,6 +34,7 @@
 #include "strlcpy.h"
 #include "util.h"
 #include "option-parser.h"
+#include "table-printer.h"
 
 #define LOG(level, fmt, ...) \
 	fprintf(stderr, "[%s:%d] <" level "> " fmt "\n", __FILE__, __LINE__, \
@@ -49,11 +51,11 @@ static bool do_debug = false;
 
 static struct cmd_info internal_events[] = {
 	{ .name = "wayvnc-startup",
-		.description = "Sent when a successful wayvnc control connection is established and event registration has succeeded, both upon initial startup and on subsequent registrations with --reconnect.",
+		.description = "Sent by wayvncctl when a successful wayvnc control connection is established and event registration has succeeded, both upon initial startup and on subsequent registrations with --reconnect.",
 		.params = {{}},
 	},
 	{ .name = "wayvnc-shutdown",
-		.description = "Sent when the wayvnc control connection is dropped, usually due to wayvnc exiting.",
+		.description = "Sent by wayvncctl when the wayvnc control connection is dropped, usually due to wayvnc exiting.",
 		.params = {{}},
 	},
 };
@@ -643,23 +645,39 @@ static int ctl_client_print_single_command(struct ctl_client* self,
 void ctl_client_print_command_list(FILE* stream)
 {
 	fprintf(stream, "Commands:\n");
+	size_t max_namelen = 0;
 	for (size_t i = 0; i < CMD_LIST_LEN; ++i) {
 		if (i == CMD_HELP) // hidden
 			continue;
-		fprintf(stream, "    %s\n", ctl_command_list[i].name);
+		max_namelen = MAX(max_namelen, strlen(ctl_command_list[i].name));
+	}
+	struct table_printer printer;
+	table_printer_init(&printer, stdout, max_namelen);
+	for (size_t i = 0; i < CMD_LIST_LEN; ++i) {
+		if (i == CMD_HELP) // hidden
+			continue;
+		table_printer_print_line(&printer, ctl_command_list[i].name,
+				ctl_command_list[i].description);
 	}
 	fprintf(stream, "\nRun 'wayvncctl command-name --help' for command-specific details.\n");
 }
 
 static void print_event_info(const struct cmd_info* info)
 {
-	printf("Event: %s\n\n%s\n", info->name,
-			info->description);
+	printf("%s\n\n", info->name);
+	table_printer_indent_and_reflow_text(stdout, info->description, 80, 0, 0);
 	if (info->params[0].name != NULL) {
-		printf("\nData fields:");
+		printf("\nData fields:\n");
+		size_t max_namelen = 0;
 		for (int i = 0; info->params[i].name != NULL; ++i)
-			printf("\n  %s=...\n    %s\n", info->params[i].name,
-					info->params[i].description);
+			max_namelen = MAX(max_namelen, strlen(info->params[i].name));
+
+		struct table_printer printer;
+		table_printer_init(&printer, stdout, max_namelen);
+		for (int i = 0; info->params[i].name != NULL; ++i)
+			table_printer_print_fmtline(&printer,
+					info->params[i].description,
+					"%s=...", info->params[i].name);
 	}
 }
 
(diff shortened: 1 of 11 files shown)


src/table-printer.c
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) 2022 Andri Yngvason, Jim Ramsay
```
any1, line 2: The year is 2023. It's better to split this into two lines.

src/ctl-client.c
```diff
@@ -41,6 +42,12 @@
 #define WARN(fmt, ...) \
 	LOG("WARNING", fmt, ##__VA_ARGS__)
 
+#define SET_MAXLEN(accumulator, string) { \
```
any1, line 45: I'm not a fan of this macro.

Consider using something like this instead:
```
accumulator = MAX(accumulator, strlen(string));
```

src/ctl-client.c
```diff
@@ -643,23 +650,48 @@ static int ctl_client_print_single_command(struct ctl_client* self,
...
+	printf("%s\n\n", info->name);
+	table_printer_indent_and_reflow_text(stdout, info->description, 76, 4, 4);
 	if (info->params[0].name != NULL) {
-		printf("\nData fields:");
+		printf("\nData fields:\n");
+		int max_namelen = 0;
 		for (int i = 0; info->params[i].name != NULL; ++i)
-			printf("\n  %s=...\n    %s\n", info->params[i].name,
-					info->params[i].description);
+			SET_MAXLEN(max_namelen, info->params[i].name);
```
any1, line 683: Please, try to add some vertical spacing where it aides readability.

src/wayvncctl.c
```diff
@@ -44,9 +44,9 @@ struct wayvncctl {
 static int wayvncctl_usage(FILE* stream, struct option_parser* options, int rc)
 {
 	static const char* usage =
-"Usage: wayvncctl [options] [command [--param1=value1 ...]]\n"
+"Usage: wayvncctl [options] <command> [parameters]\n"
 "\n"
-"Connects to and interacts with a running wayvnc instance.";
+"    Connects to and interacts with a running wayvnc instance.";
```
any1, line 49: I'm not sure that I agree that this should be indented.

src/ctl-client.c
```diff
@@ -683,10 +715,24 @@ static int print_event_details(const char* evt_name)
...
 {
 	printf("Events:\n");
+	int max_namelen = 0;
 	for (size_t i = 0; i < EVT_LIST_LEN; ++i)
-		printf("    %s\n", ctl_event_list[i].name);
+		SET_MAXLEN(max_namelen, ctl_event_list[i].name);
 	for (size_t i = 0; i < INTERNAL_EVT_LEN; ++i)
-		printf("    %s\n", internal_events[i].name);
+		SET_MAXLEN(max_namelen, internal_events[i].name);
+	struct table_printer printer = {
```
any1, line 723: This is defined the same way 3 times. `max_width`, `left_indent` and `column_offset` should be defined in one place.

src/table-printer.c
```diff
@@ -0,0 +1,102 @@
...
+	table_printer_indent_and_reflow_text(stdout,
+			right_text,
+			column_width, 0, column_indent);
+}
+
+void table_printer_print_fmtline(struct table_printer* self,
+		const char* right_text,
+		const char* left_format, ...)
+{
+	char buf[80];
```
any1, line 95: In the public interface of this class, this number is adjustable. Better make this oversized and add an assert to ensure that `sizeof(buf) > self->max_width + 2`. 256 should suffice.

any1 (changes requested):

src/table-printer.c
lack: This buffer is only for the left column, so I figured even for very wide displays, 80 chars for the left-part (command name, event name, option short+long name), 80 is pretty generous...

src/table-printer.c
```diff
@@ -0,0 +1,105 @@
...
+	int dst_len = 0;
+	int i = 0;
+
+	while (true) {
+		char c = src[i];
+		if (line_len > width) {
+			// first word > width
+			assert(last_space_pos > 0);
+			// subsequent word > width
+			assert(dst[last_space_pos] != '\n');
```
lack, line 40: Adding this assert stops an infinite loop in the unit test for width=4

src/table-printer.c
```diff
@@ -0,0 +1,105 @@
...
+
+			dst_len -= i - last_space_pos;
+			dst[dst_len++] = '\n';
+			i = last_space_pos + 1;
+			line_len = 0;
+			continue;
+		}
+		if (!c)
+			break;
```
lack, line 49: moving the loop break here prevents an off-by-one error if the end of `src` is one character longer than `width` (unit test with width=17)

lack:
@any1 : Ready for re-review any time!

Also added unit tests for the reflow and table printing code which exposed a couple edge cases (also fixed).

any1 (changes requested):

test/table-printer-test.c
```diff
@@ -0,0 +1,124 @@
...
+	return 0;
+}
+
+static int test_indent_and_reflow(void)
+{
+	size_t len;
+	char* buf;
+	FILE* stream;
+
+	stream = open_memstream(&buf, &len);
```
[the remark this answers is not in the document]
lack: Cleaner for testing, maybe, but the API becomes a chore of managing buffers then.

src/ctl-client.c
```diff
@@ -47,13 +49,23 @@ static bool do_debug = false;
 	if (do_debug) \
 	LOG("DEBUG", fmt, ##__VA_ARGS__)
 
+static struct table_printer printer = {
+	.max_width = 80,
+	.left_indent = 4,
+	.column_offset = 8,
+};
+#define INIT_PRINTER(left_width_) { \
```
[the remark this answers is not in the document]
lack: An init function makes sense... Then the option-parser could call the same function, as we would want everything to share the same defaults.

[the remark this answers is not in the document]
any1: Yes

test/table-printer-test.c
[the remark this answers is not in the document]
any1: One way to do it would be to expose a function that only does the formatting, which is then used by the function that does the printing. The formatting function would then be the one that's tested. Anyway, it was just a thought. This is good enough as is.

lack:
@any1 Ready for re-re-review ;)

src/table-printer.c
```diff
@@ -0,0 +1,135 @@
...
+
+#include "table-printer.h"
+
+#include <stdlib.h>
+#include <stdarg.h>
+#include <stdbool.h>
+#include <string.h>
+#include <assert.h>
+
+static struct table_printer defaults = {
```
any1, line 26: Nit: This could be made const

src/table-printer.c
```diff
@@ -0,0 +1,135 @@
...
+};
+
+void table_printer_set_defaults(int max_width, int left_indent, int column_offset)
+{
+	defaults.max_width = max_width;
+	defaults.left_indent = left_indent;
+	defaults.column_offset = column_offset;
+}
+
+struct table_printer* table_printer_new(FILE* stream, int left_width)
```
any1, line 41: Nit: There is no need for this to be dynamically allocated. `table_printer_init` would be enough.

any1 (approved):
LGTM

src/table-printer.c
lack: Not if I want table_printer_set_defaults to work as it is now.

Arguable if that's actually what we want, I guess ;)

src/table-printer.c
lack: Fair. And that would make the destroy unneeded.

I think I'll do that and then merge this.

PR #220, opened 2023-01-08
Positional Parameters

### LINKED ISSUE #215: For wayvncctl commands with only one parameter, make it positional
Makes using it easier.

lack:
- Allow positional args for single-param commands
- Add help output for wayvncctl positional arguments
- Add description and argument help to wayvnc --help

Fixes #215

### DIFF
--- a/src/ctl-client.c
+++ b/src/ctl-client.c
@@ -737,9 +737,12 @@ static int print_command_usage(struct ctl_client* self,
 	for (int i = 0; i < cmd_options->n_opts; ++i)
 		if (cmd_options->options[i].positional)
 			printf("<%s> ", cmd_options->options[i].positional);
+
 	printf("[parameters]\n\n");
 	table_printer_indent_and_reflow_text(stdout, info->description, 80, 0, 0);
 	printf("\n");
+	if (option_parser_print_arguments(cmd_options, stdout))
+		printf("\n");
 	option_parser_print_options(cmd_options, stdout);
 	printf("\n");
 	option_parser_print_options(parent_options, stdout);
@@ -772,6 +775,7 @@ int ctl_client_init_cmd_parser(struct option_parser* parser, enum cmd_type cmd)
 	if (param_count == 1) {
 		// Represent a single parameter as a positional argument
 		options[0].positional = info->params[0].name;
+		options[0].help = info->params[0].description;
 		i++;
 	} else {
 		for (; i < param_count; ++i) {
--- a/src/ctl-client.c
+++ b/src/ctl-client.c
@@ -733,7 +733,11 @@ static int print_command_usage(struct ctl_client* self,
 		WARN("No such command");
 		return 1;
 	}
-	printf("Usage: wayvncctl [options] %s [parameters]\n\n", info->name);
+	printf("Usage: wayvncctl [options] %s ", info->name);
+	for (int i = 0; i < cmd_options->n_opts; ++i)
+		if (cmd_options->options[i].positional)
+			printf("<%s> ", cmd_options->options[i].positional);
+	printf("[parameters]\n\n");
 	table_printer_indent_and_reflow_text(stdout, info->description, 80, 0, 0);
 	printf("\n");
 	option_parser_print_options(cmd_options, stdout);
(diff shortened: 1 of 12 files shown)


src/ctl-client.c
```diff
@@ -764,12 +771,19 @@ int ctl_client_init_cmd_parser(struct option_parser* parser, enum cmd_type cmd)
...
 	struct wv_option* options = calloc(alloc_count,
 			sizeof(struct wv_option));
-	size_t i;
-	for (i = 0; i < param_count; ++i) {
-		struct wv_option* option = &options[i];
-		option->long_opt = info->params[i].name;
-		option->help = info->params[i].description;
-		option->schema = "<value>";
+	size_t i = 0;
+	if (param_count == 1) {
```
any1, line 775: It grants us more flexibility to add a flag for this to `struct cmd_param_info`. That way we can mix positional and named options.

src/ctl-client.c
```diff
@@ -764,12 +771,19 @@ int ctl_client_init_cmd_parser(struct option_parser* parser, enum cmd_type cmd)
...
+		// Represent a single parameter as a positional argument
+		options[0].positional = info->params[0].name;
+		options[0].help = info->params[0].description;
+		i++;
+	} else {
+		for (; i < param_count; ++i) {
+			struct wv_option* option = &options[i];
+			option->long_opt = info->params[i].name;
+			option->help = info->params[i].description;
+			option->schema = "<value>";
```
any1, line 785: It might be nice the also have "schema" in `struct cmd_param_info`.

src/option-parser.c
```diff
@@ -104,6 +105,31 @@ void option_parser_print_options(struct option_parser* self, FILE* stream)
 		format_option(&printer, &self->options[i]);
 }
 
+int option_parser_print_arguments(struct option_parser* self, FILE* stream)
```
any1, line 108: It might be convenient to also generate the "Usage" line, don't you think?

src/ctl-client.c
any1: Also, in general, I favour explicit over implicit functionality.

src/ctl-client.c
lack: Yes, I was thinking the same thing.

src/option-parser.c
lack: I think that's a nice enhancement to add in the future, sure!

src/ctl-client.c
lack: Hmmm...  In some ways the cmd_param_info is more about documenting the actual JSON IPC mechanism.  How the `wayvncctl` command represents those parameter(s) as a commandline interface is a different thing...

lack: I suppose one flag that might make sense is whether a parameter is optional...

any1: As is, information about what the command line interface should look like is being derived from the IPC description anyway. I don't think that it would hurt to add some contextual hints to the IPC description.

In any case, the IPC serves the CLI, not the other way around.

lack: I guess you could see it that way!  I think of the IPC as independent, and a unique API in its own right, but I think you're right that there's no harm adding this to the command description.

any1: The alternative would be to create a separate description for the CLI that's less tightly coupled with the IPC.

lack:
@any1 Okay! The option parser now generates the usage and the command summary.  I think it's pretty good now!

src/main.c
```diff
@@ -1270,9 +1272,11 @@ int main(int argc, char* argv[])
 
 	static const struct wv_option opts[] = {
 		{ .positional = "address",
-		  .help = "The IP address or unix socket path to listen on. Default: 127.0.0.1" },
+		  .help = "The IP address or unix socket path to listen on.",
+		  .default_ = DEFAULT_ADDRESS},
 		{ .positional = "port",
-		  .help = "The TCP port to listen on. Default: 5900" },
+		  .help = "The TCP port to listen on.",
+		  .default_ =  xstr(DEFAULT_PORT)},
```
any1, line 1280: Nit: There seems to be an extra space here.

src/option-parser.c
```diff
@@ -105,6 +106,28 @@ void option_parser_print_options(struct option_parser* self, FILE* stream)
 		format_option(&printer, &self->options[i]);
 }
 
+void option_parser_print_usage(struct option_parser* self, FILE* stream)
+{
+	fputs(" [", stream);
+	for(const char* c = self->name; *c != '\0'; c++)
```
any1, line 122: Style. Also, it's better to make a function out of a code block such as this. Something like `char* string_to_lower(char* buf, const char* src)`.

src/option-parser.c
```diff
@@ -105,6 +106,28 @@ void option_parser_print_options(struct option_parser* self, FILE* stream)
...
+		fputc(tolower(*c), stream);
+	fputc(']', stream);
+	int paren_count = 0;
+	for (int i = 0; i < self->n_opts; ++i) {
+		const struct wv_option* opt = &self->options[i];
+		if (!opt->positional)
+			continue;
+		const char* paren = "";
+		if (opt->default_) {
+			paren = "[";
```
any1, line 132: Shouldn't non-optional arguments be enclosed by `<` `>`?

It might be a good idea to validate that there is no mandatory argument after an optional one, as that wouldn't work.

lack: The current code will display a required positional argument like this: `<required>` and an optional arg like this: `[<optional>]`

any1: `[<optional>]` is a bit much. I think it should be either `<required>` or `[optional]`.

lack: I added an assert about ordering... Are you okay with the usage looking like this:
```
commandname [options] <required1> <required2> [<optional1> [<optional2>]]
```

any1: I prefer the latter, but it's a cosmetic thing so I don't really care a lot about it. We can change it later.

lack: While I'm in there I can gladly change it :)

lack: Done!

Issue #232, opened 2023-01-27
Build failure: wayland-client.h: No such file or directory 


jubalh:
Since 0.6.0 (or 0.6.1) I get the following build failure:
```
[   10s] ../include/output.h:19:10: fatal error: wayland-client.h: No such file or directory
```
This is even though meson finds the dependency:
```
[   10s] Run-time dependency wayland-client found: YES 1.21.0
```
See whole build log: https://build.opensuse.org/build/X11:Wayland/openSUSE_Tumbleweed/x86_64/wayvnc/_log

This is also true for 0.6.2.

any1:
Did someone remove that header from Suse's libwayland package?

jubalh:
No, the header is installed at: `/usr/include/wayland/wayland-client.h`

any1:
Looks more like a problem with your distro's pkg-config script for libwayland. In any case, not a wayvnc issue as far as I can tell.

jubalh:
> Looks more like a problem with your distro's pkg-config script for libwayland.

In all other programs that use wayland-client.h, like sway or cage, it works without issues.

jubalh:
@emersion I think you know your way well around meson and wayland things. Do you by any chance spot the issue?

emersion:
Maybe `wayvncctl` sources include this header by mistake? wayland-client is missing from `wayvncctl` dependencies (for good reason).

any1:
> Maybe wayvncctl sources include this header by mistake?

Yeah, ctl-client.c includes ctl-server.h for some reason...

jubalh:
It seems to be that actually nothing in `ctl-client.c` uses things from `ctl-server.h`. It builds fine without that includes and I don't see any warnings nor errors. Should I open a PR?

PR #234, opened 2023-01-27
Dont include ctrl-server.h in client file

jubalh:
Seems to be a copy/paste mistake.
I don't see that any of the definitions of that header are used in `ctl-client.c`.

Fix https://github.com/any1/wayvnc/issues/232


any1:
Yep, that seems fine.

any1:
Thanks!

PR #235, opened 2023-01-28
Add initial integration test to CI

lack:
The integration test spins up a headless sway, and wayvnc, and exercises
a handful of wayvncctl commands to ensure both wayvnc and wayvncctl
retain basic functionality.

This change also enables the integration test on all 3 CI runners.


any1:
Probably good enough, except we should use tabs for indenting bash scripts for the sake of consistency.

Also, we might want to put an "unlicense" notice on tests and examples so people can use them however they want.

PR #239, opened 2023-02-07
main: Fix unix-socket option parsing

Waldteufel:
option_parser_get_value was looking for "unix" instead of "unix-socket" before, so it was impossible to switch to unix socket mode


any1:
Thanks!

PR #252, opened 2023-05-31
Fix wrong version number

Enzime:


any1:
Right. It seems that I forgot to increase the version number on last release, but this isn't going to help with that. Besides, v0.6 is on it's own branch, not master.

PR #261, opened 2023-09-26
get pixel size from fourcc to calculate pixel stride

maztheman:
implementing the TODO in buffer.c

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -92,6 +92,7 @@ struct wv_buffer* wv_buffer_create_shm(int width,
 		goto shm_failure;
 
 	int bpp = pixel_size_from_fourcc(fourcc);
+	assert(bpp > 0);
 	self->nvnc_fb = nvnc_fb_from_buffer(self->pixels, width, height, fourcc,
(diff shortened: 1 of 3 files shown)


maztheman:
vncdo didnt exist in archlinux...hmm odd.

src/buffer.c
```diff
@@ -91,9 +91,9 @@ struct wv_buffer* wv_buffer_create_shm(int width,
 	if (!self->wl_buffer)
 		goto shm_failure;
 
-	// TODO: Get the pixel size from the format instead of assuming it's 4.
+	int bpp = pixel_size_from_fourcc(fourcc);
 	self->nvnc_fb = nvnc_fb_from_buffer(self->pixels, width, height, fourcc,
-			stride / 4);
+			stride / bpp);
```
Consolatis, line 96: Just passing by: this divides by zero if no format matches.

maztheman: Suppose an assert(bpp > 0); could be added.

any1:
Merged externally. Thanks!

Issue #57, opened 2020-07-19
Client side desktop resizing


any1:
When the selected output is a headless output, it should be possible to resize the output according to the client's wishes. This should be turned on using a command line flag or a config variable.

any1:
Completed in #270

derekschrock:
Where did the command line option go or was it never added?   It seem like it would be good to have an option to disable resizing.

PR #270, opened 2023-10-26
Add support for dynamic resizing of headless outputs based on VNC client size changes

Consolatis:
This is just an initial draft to discuss the overall design of the feature.
Fixes:
- #57 (for headless outputs at least)

Todo:
- [x] `When declaring pointer variables, the asterisk (*) is placed on the left with the type rather than the variable name.`
- [x] `User-visible string such as log messages must not be split up`
- [x] ~~Move `output->is_headless` to `Automatically resize headless outputs on client request` commit~~
- [x] Remove the `[wip]`
- [ ] Test display hotplug
- [x] Squash the fixup commits

Also: /me likes brown M&Ms.

### DIFF
--- a/include/output-management.h
+++ b/include/output-management.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2023 The wayvnc authors
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+ * REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+ * INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+ * LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE
+ * OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#pragma once
+
+#include <stdint.h>
+#include <stdbool.h>
+
+struct output;
+struct zwlr_output_manager_v1;
+
+void wlr_output_manager_setup(struct zwlr_output_manager_v1* output_manager);
+bool wlr_output_manager_resize_output(struct output* output,
+	uint16_t width, uint16_t height);
+void wlr_output_manager_destroy(void);
--- a/src/main.c
+++ b/src/main.c
@@ -39,12 +39,14 @@
 #include "virtual-keyboard-unstable-v1.h"
 #include "xdg-output-unstable-v1.h"
 #include "wlr-output-power-management-unstable-v1.h"
+#include "wlr-output-management-unstable-v1.h"
 #include "linux-dmabuf-unstable-v1.h"
 #include "ext-transient-seat-v1.h"
 #include "screencopy.h"
 #include "data-control.h"
 #include "strlcpy.h"
 #include "output.h"
+#include "output-management.h"
 #include "pointer.h"
 #include "keyboard.h"
 #include "seat.h"
@@ -248,6 +250,15 @@ static void registry_add(void* data, struct wl_registry* registry,
 		return;
 	}
 
+	if (strcmp(interface, zwlr_output_manager_v1_interface.name) == 0) {
+		nvnc_trace("Registering new wlr_output_manager");
+		struct zwlr_output_manager_v1* wlr_output_manager =
+			wl_registry_bind(registry, id,
+				&zwlr_output_manager_v1_interface, 1);
+		wlr_output_manager_setup(wlr_output_manager);
+		return;
+	}
+
 	if (strcmp(interface, zwlr_screencopy_manager_v1_interface.name) == 0) {
 		self->screencopy.manager =
 			wl_registry_bind(registry, id,
@@ -372,6 +383,8 @@ void wayvnc_destroy(struct wayvnc* self)
 
 	if (wlr_output_power_manager)
 		zwlr_output_power_manager_v1_destroy(wlr_output_power_manager);
+
+	wlr_output_manager_destroy();
 
 	wl_shm_destroy(wl_shm);
 
--- a/src/main.c
+++ b/src/main.c
@@ -681,6 +681,34 @@ static void on_client_cut_text(struct nvnc_client* nvnc_client,
 	}
 }
 
+static bool on_client_resize(struct nvnc_client* nvnc_client,
+		const struct nvnc_desktop_layout* layout)
+{
+	struct wayvnc_client* client = nvnc_get_userdata(nvnc_client);
+
+	uint16_t width = nvnc_desktop_layout_get_width(layout);
+	uint16_t height = nvnc_desktop_layout_get_height(layout);
+	struct output* output = client->server->selected_output;
+
+	if (output == NULL)
+		return false;
+
+	nvnc_log(NVNC_LOG_DEBUG,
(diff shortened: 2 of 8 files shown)


any1:
The over all approach seems fine to me.

src/main.c
```diff
@@ -668,6 +681,33 @@ static void on_client_cut_text(struct nvnc_client* nvnc_client,
...
+
+	if (output == NULL)
+		return false;
+
+	nvnc_log(NVNC_LOG_DEBUG, "Client resolution changed: %ux%u, "
+		"capturing output %s which is headless: %s", width, height,
+		output->name, output->is_headless ? "yes" : "no");
+
+	/*
+	 * TODO: neatvnc currently does not support anything other than
```
any1, line 701: Yes, at a glance, it seems that neatvnc doesn't even perform this action synchronously under any circumstances. It will always send a second update as soon as it receives a framebuffer with different dimensions. It's probably best to just change neatvnc's response to `RFB_RESIZE_STATUS_REQUEST_FORWARDED`.

Consolatis:
Fixed the first codestyle issue with pointers on the left.
There are some more in the codebase (found via ` grep -Rn '. [*][^/ ]' src/`):
```
src/main.c:335:static int find_render_node(char *node, size_t maxlen) {
src/main.c:337: drmDevice *devices[64];
src/main.c:341:         drmDevice *dev = devices[i];
src/main.c:538:static struct ctl_server_client *client_next(struct ctl* ctl,
src/main.c:539:         struct ctl_server_client *prev)
src/main.c:551: const struct nvnc_client *vnc_client =
src/main.c:553: const struct wayvnc_client *client = nvnc_get_userdata(vnc_client);
src/main.c:572: struct ctl_server_output* item = *outputs;
src/output.c:272:static void output_power_mode(void *data,
src/output.c:273:                    struct zwlr_output_power_v1 *zwlr_output_power_v1,
src/output.c:293:static void output_power_failed(void *data,
src/output.c:294:                    struct zwlr_output_power_v1 *zwlr_output_power_v1)
src/strlcpy.c:28:strlcpy(char *dst, const char *src, size_t dsize)
src/strlcpy.c:30:       const char *osrc = src;
src/shm.c:29:static inline int memfd_create(const char *name, unsigned int flags) {
src/shm.c:35:static void randname(char *buf)
src/option-parser.c:119:static void print_string_tolower(FILE* stream, const char *src)
src/ctl-server.c:862:           const struct ctl_server_client_info *info,
src/ctl-server.c:915:           const struct ctl_server_client_info *info,
src/ctl-server.c:924:           const struct ctl_server_client_info *info,
src/ctl-server.c:932:           const struct ctl_server_client_info *info,
```

Consolatis:
This PR is ready from my point of view. Will obviously address further review comments.

include/output-management.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <stdint.h>
+#include <stdbool.h>
+
+struct output;
+struct zwlr_output_manager_v1;
+
+void wlr_output_manager_setup(struct zwlr_output_manager_v1* output_manager);
+bool wlr_output_manager_resize_output(struct output *output,
```
any1, line 26: whoops

src/output-management.c
```diff
@@ -0,0 +1,265 @@
...
+
+static void output_head_finished(void* data,
+		struct zwlr_output_head_v1* output_head)
+{
+	nvnc_trace("head gone, removing");
+	struct output_manager_head* head = data;
+	zwlr_output_head_v1_destroy(output_head);
+	wl_list_remove(&head->link);
+	free(head->name);
+	head->name = NULL;
```
any1, line 110: Nit: Setting `NULL` here is somewhat superfluous if you're calling `free` right after.

src/output-management.c
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	if (!output->is_headless) {
+		nvnc_log(NVNC_LOG_INFO,
+			"not resizing output %s: not a headless one",
+			output->name);
+		return false;
+	}
+
+	// TODO: This could be synced to --max-fps
```
any1, line 233: Yeah, that would actually be a good idea (for later). There might be some impedance mismatch though because the headless refresh timer has a millisecond resolution while wayvnc's timer has microsecond resolution. So we'll end up with one timer having a 16.000 µs interval and the other 16.667 µs.

It might be a good idea to fix that in wlroots. There might even be a PR for it already.

src/output-management.c
```diff
@@ -0,0 +1,265 @@
...
+		if (!head->enabled) {
+			nvnc_trace("disabling output %s", head->name);
+			zwlr_output_configuration_v1_disable_head(
+				config, head->head);
+			continue;
+		}
+
+		config_head = zwlr_output_configuration_v1_enable_head(
+			config, head->head);
+		if (head->name && !strcmp(head->name, output->name)) {
```
any1, line 255: Nit: Prefer `strcmp() == 0`

src/output.c
```diff
@@ -219,7 +219,10 @@ void output_name(void* data, struct zxdg_output_v1* xdg_output,
 	struct output* self = data;
 
 	strlcpy(self->name, name, sizeof(self->name));
-	nvnc_trace("Output %u name: %s", self->id, self->name);
+	self->is_headless = !strncmp(name, "HEADLESS-", strlen("HEADLESS-"));
```
any1, line 222: Ditto

src/main.c
any1: I think we should turn this TODO into a Neat VNC issue.

any1 (commented):
output-manager is notoriously finicky and tends to crash due to object id re-use.

Have you tried adding and removing headless outputs while running? Perhaps event disconnecting and connecting a physical display?

src/output-management.c
any1: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3739

include/output-management.h
Consolatis: whoops indeed. Only checked the `src/` directory but completely forgot about the header.

src/output-management.c
Consolatis: True. I tend to do it anyway because dereferencing a `NULL` pointer crashes way louder than dereferencing a random memory address in case of a use after free. Can remove the two `NULL` assignments though if preferred.

Consolatis:
> output-manager is notoriously finicky and tends to crash due to object id re-use.
> 
> Have you tried adding and removing headless outputs while running? Perhaps event disconnecting and connecting a physical display?

No, in fact I didn't. Will test when I get some time.
In the meantime I added fixup commits for the two code style issues, will squash them when I confirmed outputs being removed / added works correctly.

any1:
It's fine if you don't have time for though testing. We can just push things to master as-is and people who use the master branch might report crashes if they run into them.

By the way, what happens if multiple clients with resizing capability connect? Won't they start fighting?

Consolatis:
> It's fine if you don't have time for though testing.

Main issue for this is that my daily driver and my test system both only have a single HDMI output. So I need to setup an old laptop for testing it.

> We can just push things to master as-is and people who use the master branch might report crashes if they run into them.

That is obviously your decision as the maintainer of the project. I'll squash the commits later today.

> By the way, what happens if multiple clients with resizing capability connect? Won't they start fighting?

I didn't test that but I'd assume that would only happen if they react to a resized framebuffer by enforcing their current window size again rather than using scaling or adjusting their window size. I am not sure how wayvnc could prevent that scenario though. (Although we could try to prevent resizing if the size is already set to the target size, but that is more of a small generic optimization as I *think* we wouldn't send a new size to the clients if the size didn't actually change).

Just a heads up: I also did not test output scaling or transform and how that may affect the "physical hardware units of the output device" when using `set_custom_mode()`.

any1:
> Just a heads up: I also did not test output scaling or transform and how that may affect the "physical hardware units of the output device" when using set_custom_mode().

Scaling is probably not a problem, but 90 or 270 degree rotation will swap width and height. That's why we have `output_get_transformed_width()` and `output_get_transformed_height()`. But, it doesn't make sense for a headless output to be rotated, because wayvnc is just going to rotate the image back anyway. Maybe we should just force 0 degree rotation when a client sets the resolution.

Perhaps the simplest way to keep clients from fighting would be to only enable dynamic resizing when a single client is connected. If you connect while another is connected and you want to resize, then you can just kick the other client out by not setting the "shared" flag when connecting.

src/main.c
any1: https://github.com/any1/neatvnc/issues/106

any1:
Merged. Thanks!

PR #306, opened 2024-04-19
meson: Specify check arg in run_command()

emersion:
Fixes the following warning:

    WARNING: You should add the boolean check kwarg to the run_command call.
             It currently defaults to false,
             but it will default to true in future releases of meson.
             See also: https://github.com/mesonbuild/meson/issues/9300


any1:
Thanks!

PR #309, opened 2024-05-21
main: Added quit on last client disconnect

cybin:
I needed an option to automatically quit wayvnc when the last client disconnect.

Added a condition in client_destroy() and set self->server->do_exit to true, if condition is met.

For example, if wayvnc will only run intentionally by starting via SSH, it automatically
quit when I close the VNC connection and no other clients are connected.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1726,7 +1726,6 @@ int main(int argc, char* argv[])
 	bool use_transient_seat = false;
 
 	bool quit_on_disconnect = false;
-	bool first_client_connected = false;
 
 	int drm_fd MAYBE_UNUSED = -1;
 
--- a/src/main.c
+++ b/src/main.c
@@ -84,7 +84,7 @@ enum socket_type {
 
 struct wayvnc {
 	bool do_exit;
-	bool quit_on_disconnect;
+	bool exit_on_disconnect;
 
 	struct wl_display* display;
 	struct wl_registry* registry;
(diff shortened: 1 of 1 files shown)


any1:
Hi, have you read CONTRIBUTING.md?

src/main.c
```diff
@@ -1296,6 +1297,10 @@ static void client_destroy(void* obj)
 	nvnc_log(NVNC_LOG_DEBUG, "Client disconnected, new client count: %d",
 			wayvnc->nr_clients);
 
+	if (self->server->quit_on_disconnect && self->server->nr_clients == 0) {
+		self->server->do_exit = true;
```
any1, line 1301: Nit: Prefer calling `wayvnc_exit` here.

src/main.c
```diff
@@ -1752,6 +1759,8 @@ int main(int argc, char* argv[])
 		  .default_ = "30" },
 		{ 'p', "performance", NULL,
 		  "Show performance counters." },
+		{ 'q', "quit", NULL,
+			"Quit on last client disconnect" },
```
any1, line 1763: Nit: This should be indented with tabs and aligned with spaces.

src/main.c
```diff
@@ -1752,6 +1759,8 @@ int main(int argc, char* argv[])
 		  .default_ = "30" },
 		{ 'p', "performance", NULL,
 		  "Show performance counters." },
+		{ 'q', "quit", NULL,
```
any1, line 1762: Let's call this `quit-on-disconnect` or even `exit-on-disconnect`, as the system call is called `exit` after all.

cybin:
Thanks for your help.

So, I'm going to rename everything to exit-on-disconnect and change 'q' to 'e' accordingly. Will push later today, when I got the time to test it.

any1:
LGTM. Please squash into single commit. Keep in mind that commit messages should be written in imperative mood. A commit message should complete the sentence "When applied, this patch will...".

I.e. "main: Added quit on last client disconnect" might be written "main: Add option to exit on last disconnect" instead.

any1:
Pulled

Issue #227, opened 2023-01-18
External listener fd


any1:
There are a few different types of sockets available which people might want to use for listening to VNC connections. This includes "vsock", websockets and systemd sockets.

In order to accommodate "any kind of socket", we can make it possible to pass a socket from a parent process to wayvnc to be used with `listen()`. See https://gitlab.freedesktop.org/wayland/weston/-/commit/2f9319cef611a1c2925e65d63e1068b8f80a531b for reference.

This issue supersedes the following issues and pull requests
 * #144
 * #145 
 * #146

PR #310, opened 2024-05-31
External listener fd

layercak3:
Tested with a systemd .socket unit.

Closes: https://github.com/any1/wayvnc/issues/227
Requires: https://github.com/any1/neatvnc/pull/111

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -80,6 +80,7 @@ enum socket_type {
 	SOCKET_TYPE_TCP = 0,
 	SOCKET_TYPE_UNIX,
 	SOCKET_TYPE_WEBSOCKET,
+	SOCKET_TYPE_FROM_FD,
 };
 
 struct wayvnc {
@@ -885,7 +886,7 @@ static char* get_cfg_path(const struct cfg* cfg, char* dst, const char* src)
 }
 
 static int init_nvnc(struct wayvnc* self, const char* addr, uint16_t port,
-		enum socket_type socket_type)
+		int fd, enum socket_type socket_type)
 {
 	switch (socket_type) {
 		case SOCKET_TYPE_TCP:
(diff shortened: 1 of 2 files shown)


layercak3:
Looks like the sr.ht CI is being activated on GitHub PRs somehow but I don't see any indication of it checking out to my commit object. If it needs to be on PRs the linker should be failing on mine.

src/main.c
```diff
@@ -1705,6 +1711,8 @@ int main(int argc, char* argv[])
 
 	const char* address = NULL;
 	int port = 0;
+	bool use_external_listener_fd = false;
```
any1, line 1714: For other socket types, we use boolean command line options to select which kind of "address" we have in the address positional argument. This diverges from that convention.

src/main.c
```diff
@@ -1754,6 +1762,8 @@ int main(int argc, char* argv[])
 		  "Show performance counters." },
 		{ 'u', "unix-socket", NULL,
 		  "Create unix domain socket." },
+		{ 'e', "external-listener-fd", "<fd>",
+		  "Have the VNC server listen on this socket fd, instead of binding to an address and listening on that." },
```
any1, line 1766: I like to keep these explanations as short as possible. Can you think of a way to make this shorter?

src/main.c
```diff
@@ -1831,13 +1845,23 @@ int main(int argc, char* argv[])
...
+	}
+
 	if (use_transient_seat && disable_input) {
 		nvnc_log(NVNC_LOG_ERROR, "transient-seat and disable-input are conflicting options");
 		return 1;
 	}
 
 	if (seat_name && use_transient_seat) {
-		nvnc_log(NVNC_LOG_ERROR, "transient seat and seat are conflicting options");
+		nvnc_log(NVNC_LOG_ERROR, "transient-seat and seat are conflicting options");
```
any1, line 1864: This would belong to a separate commit

wayvnc.scd
```diff
@@ -13,7 +13,7 @@ wayvnc - A VNC server for wlroots based Wayland compositors.
 *-C, --config=<path>*
 	Select a config file.
 
-*-g,--gpu*
+*-g, --gpu*
```
any1, line 16: Formatting changes really should be in a separate commit. If you have sufficient git skills, it would be nice if you could split the commit. Otherwise, just keep this in mind for future contributions.

src/main.c
layercak3: Its only purpose was to substitute for `external_listener_fd >= 0`, so I've removed it.

src/main.c
layercak3: Changed to "Listen on a bound socket at \<fd\> instead of binding to an address."

layercak3:
Force-pushed new commits with review addressed.

src/main.c
any1: I actually meant that the fd should just be treated as an "address". However, at some point I want to add the ability to listen on multiple addresses, so we'll need a new scheme for listing addresses.

any1:
For some reason, the integration test is segfaulting on the github CI, but not srht.

layercak3:
I deployed this on my server and external listener fd works, but it seems that I broke normal address listening after the review (I think I remember it working while working on the patch initially), maybe due to the atoi/option parser retval after removing the boolean. I'll have to fix that.
As for sr.ht, I stated earlier that I don't see any indication of it checking out to PR heads.

layercak3:
I checked on another user's PR and it was doing git checkout (https://builds.sr.ht/~andri/job/1227912), so I was wrong about that, it's only happening to my PRs, probably cause I named the repo "wayvnc-ghfork", maybe a bug in hottub-bridge :)

any1:
LGTM. There are conflicts. Can you resolve them?

any1:
Thanks!

PR #312, opened 2024-06-02
main: Use info.address_storage to compose client addr strings

layercak3:
Fixes a bug where wayvncctl was not showing the last 16 digits of my IPv6 address.


any1:
Thanks!

PR #333, opened 2024-09-12
data-control: Ignore our own offers via custom MIME type

layercak3:
It behaves as intended when I tested. I don't think the custom MIME type will interfere with anything considering password managers add `x-kde-passwordManagerHint` and Chromium/Firefox add stuff too.

### DIFF
--- a/src/data-control.c
+++ b/src/data-control.c
@@ -26,6 +26,8 @@
 
 #include "data-control.h"
 
+static const char custom_mime_type_data[] = "wayvnc";
+
 struct receive_context {
 	struct data_control* data_control;
 	struct zwlr_data_control_offer_v1* offer;
@@ -124,11 +126,15 @@ static void data_control_offer(void* data,
 {
 	struct data_control* self = data;
 
-	if (self->offer)
-		return;
-	if (strcmp(mime_type, self->mime_type) != 0) {
+	if (strcmp(mime_type, self->custom_mime_type_name) == 0) {
+		self->is_own_offer = true;
 		return;
 	}
+
+	if (self->offer)
+		return;
+	if (strcmp(mime_type, self->mime_type) != 0)
+		return;
 
 	self->offer = zwlr_data_control_offer_v1;
 }
@@ -153,9 +159,13 @@ static void data_control_device_selection(void* data,
 {
 	struct data_control* self = data;
 	if (id && self->offer == id) {
-		receive_data(data, id);
+		if (!self->is_own_offer)
+			receive_data(data, id);
+		else
+			zwlr_data_control_offer_v1_destroy(self->offer);
 		self->offer = NULL;
 	}
+	self->is_own_offer = false;
 }
 
 static void data_control_device_finished(void* data,
@@ -170,10 +180,13 @@ static void data_control_device_primary_selection(void* data,
 {
 	struct data_control* self = data;
 	if (id && self->offer == id) {
-		receive_data(data, id);
+		if (!self->is_own_offer)
+			receive_data(data, id);
+		else
+			zwlr_data_control_offer_v1_destroy(self->offer);
 		self->offer = NULL;
-		return;
 	}
+	self->is_own_offer = false;
 }
 
 static struct zwlr_data_control_device_v1_listener data_control_device_listener = {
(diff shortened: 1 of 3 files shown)


src/data-control.c
```diff
@@ -252,9 +271,14 @@ void data_control_init(struct data_control* self, struct wl_display* wl_display,
 	zwlr_data_control_device_v1_add_listener(self->device, &data_control_device_listener, self);
 	self->selection = NULL;
 	self->primary_selection = NULL;
+	self->offer = NULL;
+	self->offer_is_our_own = false;
 	self->cb_data = NULL;
 	self->cb_len = 0;
 	self->mime_type = "text/plain;charset=utf-8";
+	snprintf(self->custom_mime_type_name,
```
any1, line 279: How did you arrive at this particular number of digits for the random number?

any1 (commented):
Did you try this method suggested by emersion? https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/issues/111#note_1143403

layercak3:
> Did you try this method suggested by emersion? https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/issues/111#note_1143403

I had the assumption that anonymous pipes did not have inode numbers, but after some research, looking at /proc/$pid/fd/, and inserting fstat/log calls I was clearly wrong. They do have inode numbers, just in pipefs, not on a userspace-visible filesystem.
It is a viable solution and I can switch the PR to it if you'd like. A transfer has to be performed but the code is more simple.

any1:
> It is a viable solution and I can switch the PR to it if you'd like. A transfer has to be performed but the code is more simple.

I do value simplicity. ;)

The custom mime type might be more robust though, and portable. For example, if the other client decides to use `socketpair` with a tcp socket for some reason, that won't have an inode. A unix stream socket will have an inode, so that also works.

I suppose we could just fall back to the old behaviour if there is no inode and emit a warning message saying that the other end used a weird socket type, so let's do the inode thing, but I'd still like to keep this PR in case we change our minds later.

layercak3:
If I add a `ino_t last_send_inode;` to `struct data_control`, set it to 0 in `data_control_init` (well it's calloc'd anyway), set it to `sb.st_ino` in `data_control_source_send` (if the fd doesn't have an inode then it gets set to 0, but that also means it's definitely not our pipe write fd), and don't run `nvnc_send_cut_text` if `OurPipeReadFdStatBuf.st_ino == ctx->data_control->last_send_inode`, it doesn't completely work. While before the text would be echoed back twice (once for selection, another for primary selection), now it usually only echoes it back once.

We set both the selection and primary selection in `data_control_to_clipboard`, then we get both offers/selection events and create a receive context for both selection events.
Problem is that typically the order of events is that we handle both send events (the send callback sets `self->last_send_inode`), and only then does our event loop call `on_receive`. By the time the first `on_receive` for the first receive context is run we already overwrote `last_send_inode` with the second inode, so it won't match and the clipboard data will be echoed. Then on the second `on_receive` it will match and not echo.
Even if wayvnc didn't set the primary selection, if we were very busy with other clients asking for the clipboard many times, the issue might arise.

Unless there is a simple way to avoid that which I missed, the inode method isn't actually viable.

src/data-control.c
layercak3: Arbitrarily. I guess it's overkill and I can reduce it if you want.

any1: Nah, overkill is good in this case. In fact, I think that it would be even better to use 8 hexadecimal digits than 9 decimal digits.

any1:
Oh, well, thanks for trying out the inode thing anyway. Let's continue with the MIME thing.

src/data-control.c
layercak3: Changed it to `x-wayvnc-client-%08x", rand() % 0xffffffff`

any1: Modulo is not needed. Also, `9 % 9 = 0`. If you want a range from 0 to 9, you need to do `x % 10`.  If rand had a 64 bit return value and you wanted to make it 32 bits, the best way to do it would be to use `x & 0xffffffff` or `x & UINT_MAX` or just downcasting it to `uint32_t`.

Anyway, this should just say `(unsigned int)rand()`.

soreau:
> LGTM. @soreau, can you give this a test for us?

I read through the patch and it looks ok but even after reading the commit message, it's not clear to me what I should be testing for, exactly.

any1:
Thanks!

PR #336, opened 2024-09-17
data-control: Fix use-after-free in on_receive

layercak3:
If a client is closed between the time that one of its receive contexts is created and the time that the receive context is finished reading from its fd, when it is finished and finally calls nvnc_send_cut_text it will access ctx->data_control->server, but data_control is part of the wayvnc_client object which was freed when the client was closed.

The only purpose of the data_control pointer being in receive_context was to pass data_control->server to nvnc_send_cut_text, so receive_context can just store that server pointer itself.

This was the issue I found in https://github.com/any1/neatvnc/pull/114#issuecomment-2303229294
Can be reproduced by having a VNC client open, then `wl-copy < /some/obnoxiously/large/uncompressible/text` on the server and immediately closing the VNC client after running that command.

### DIFF
--- a/src/data-control.c
+++ b/src/data-control.c
@@ -29,7 +29,7 @@
 static const char custom_mime_type_data[] = "wayvnc";
 
 struct receive_context {
-	struct data_control* data_control;
+	struct nvnc* server;
 	struct aml_handler* handler;
 	LIST_ENTRY(receive_context) link;
 	struct zwlr_data_control_offer_v1* offer;
@@ -71,8 +71,7 @@ static void on_receive(void* handler)
 	ctx->mem_fp = NULL;
 
 	if (ctx->mem_size)
-		nvnc_send_cut_text(ctx->data_control->server, ctx->mem_data,
-				ctx->mem_size);
+		nvnc_send_cut_text(ctx->server, ctx->mem_data, ctx->mem_size);
 
 	destroy_receive_context(ctx);
 }
@@ -103,7 +102,7 @@ static void receive_data(void* data,
 	close(pipe_fd[1]);
 
 	ctx->fd = pipe_fd[0];
-	ctx->data_control = self;
+	ctx->server = self->server;
 	ctx->offer = offer;
 	ctx->mem_fp = open_memstream(&ctx->mem_data, &ctx->mem_size);
 	if (!ctx->mem_fp) {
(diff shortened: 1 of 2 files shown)


any1:
Thanks for tracking this down!

The `aml_handle` is responsible for cleaning up the context, and it shouldn't call `on_receive` after it's cleaned up, so this looks more like an aml bug.

any1:
Actually, the handler should be referenced by `data_control` and stopped and unreffed in `data_control_destroy`.

layercak3:
> Actually, the handler should be referenced by `data_control` and stopped and unreffed in `data_control_destroy`.

That makes sense. There can be multiple handlers running concurrently so they would need to be stored in some list/queue though.

I also think it that it makes sense that the receive context is decoupled from the client. I thought that was the intention when reading the code, and none of the things in receive_context are necessarily tied to the client (cleaned up when it's gone).

layercak3:
I just added the receive context/aml handler to a linked list so I can stop the receive contexts when closing the client.
I'll properly check this and test again in some hours.
I'm not familiar with aml and need to learn this but right now I think I need to `aml_unref` too (probably in `on_receive` too)

src/data-control.c
```diff
@@ -116,6 +108,9 @@ static void receive_data(void* data,
 		return;
 	}
 
+	ctx->handler = handler;
```
any1, line 110: Objects that reference the handler should call `aml_ref` (and `aml_unref` on destroy). It's not strictly needed for the over-all picture here, but it should be done for sake of consistency.

There is a slight problem with keeping a reference here though in that this will cause a circular reference. What we should do here is turn around the ownership, so that the context owns the handler. This means removing the cleanup callback from `aml_handler_new` and destroying the context in `on_receive` instead.

src/data-control.c
```diff
@@ -293,6 +289,11 @@ void data_control_destroy(struct data_control* self)
 	}
 	zwlr_data_control_device_v1_destroy(self->device);
 	free(self->cb_data);
+
+	struct receive_context* receive_context;
+	struct receive_context* tmp;
+	LIST_FOREACH_SAFE (receive_context, &self->receive_contexts, link, tmp)
```
any1, line 294: Nit: It's generally a more robust approach to use `while (!LIST_EMPTY(...))`.

layercak3:
I understand it better now. Thanks for your help and blog post.

any1:
blog post?

layercak3:
https://andri.yngvason.is/managing-object-lifetimes-in-c.html
It's pretty basic stuff but I don't have much experience :p

any1:
LGTM.

any1:
Thanks!

PR #338, opened 2024-09-27
buffer: Fix compilation without ENABLE_SCREENCOPY_DMABUF

onluen:
`wv_gbm_device` is used unconditionally in `buffer.c`, which breaks compilation. I exposed `wv_gbm_device` unconditionally and limit it's contents instead.


any1:
Thanks. `wv_create_buffer`, didn't really need to be exposed in the header, so I removed it and fixed this.

PR #345, opened 2024-10-21
actions: Fixed node16 deprecation error causing build failure

guylamar2006:
I have read and understood CONTRIBUTING.md


any1:
Nit: commit messages should be in imperative mood. I.e. "Fixed" should be "Fix".

any1:
Applied

PR #385, opened 2025-06-10
ctl: Add command to set the desktop name

layercak3:
This can be used to change the client's window title at runtime to include something like the title of the currently focused window or the representation of the focused workspace's layout.


any1:
Looks like I forgot to merge this a while back. Thanks!

PR #397, opened 2025-08-15
main: Allow websockets to use fd and unix sockets

tsukasa-au:
The docs for the `--unix-socket` and `--external-listener-fd` options sound like they should work with the `--websocket` option. This change implements that.

By allowing wayvnc to accept an arbitrary FD to listen on, we can use systemd's socket activation functionality to start wayvnc on demand.

I am adding this as I want to put wayvnc behind nginx, and only start wayvnc when a webclient is trying to connect to the backend.

--
`I have read and understood CONTRIBUTING.md`

NOTE: I was not exactly clear on how you prefer lines to be wrapped, I believe I have followed style guide, but am not certain.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -80,7 +80,6 @@ struct wayvnc_client;
 enum socket_type {
 	SOCKET_TYPE_TCP = 0,
 	SOCKET_TYPE_UNIX,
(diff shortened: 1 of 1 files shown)


any1:
Thanks!

PR #456, opened 2026-09-19
Allow disabling clipboard synchronization independently of input

krscott:
Add `--disable-clipboard` to turn off clipboard and primary-selection synchronization without disabling keyboard or pointer input. This skips per-client data-control device creation; applications on the compositor can still copy and paste normally. The default behavior is unchanged.

This is useful for input-only clients that do not need clipboard access. It also avoids a reproducible disconnect race: with an active selection, rapidly connecting and disconnecting RFB clients can leave pending `data_offer` events addressed to destroyed data-control devices. Libwayland discards those events and later rejects a new offer ID with `not a valid new object id`, disconnecting wayvnc from the compositor.

Verified by building the patch against 0.10.1 with its Meson tests and checking the new CLI option. Headless Sway integration verification covers client churn, keyboard/pointer input, and application copy/paste with forwarding disabled.


krscott:
🤖 gpt-6 on behalf of krscott:

Arch Linux and FreeBSD CI passed. The [Ubuntu job](https://builds.sr.ht/~andri/job/1891978) failed before checkout or compilation: SourceHut could not boot the configured image and reported `Missing base image functions 'ubuntu/plucky'`.

Local verification against 0.10.1 passed the Meson tests and a 2,000-connection RFB disconnect probe with clipboard forwarding disabled. The same probe crashes without the option. Application copy/paste, keyboard input, and pointer input still work with the option enabled.

any1:
Please read CONTRIBUTING.md

krscott:
apologies. meant to limit this PR to my fork, not push upstream.

